PR #172, opened 2022-11-21
Sum stats instead of last

Meldiron:
## What does this PR do?

Instead of showing last 1H/1D stat in overview, we sum all results.

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

Yes

### DIFF
--- a/src/routes/console/project-[project]/overview/+layout.svelte
+++ b/src/routes/console/project-[project]/overview/+layout.svelte
@@ -65,7 +65,7 @@
             <Onboard {projectId} />
         {:else}
             {#if $usage}
-                {@const storage = humanFileSize(last($usage.storage).value)}
+                {@const storage = humanFileSize(total($usage.storage))}
                 <section class="common-section">
                     <div class="grid-dashboard-1s-2m-6l">
                         <div class="card is-2-columns-medium-screen is-3-columns-large-screen">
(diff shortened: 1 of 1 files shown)


Meldiron:
Before:
CleanShot 2022-11-21 at 11 03 34

After:

CleanShot 2022-11-21 at 11 03 25

src/routes/console/project-[project]/overview/+layout.svelte
```diff
@@ -89,14 +89,14 @@
 
                                 <div class="grid-item-1-end-start">
                                     <div class="heading-level-4">
-                                        {format(last($usage.documents).value)}
+                                        {format(total($usage.documents))}
```
Comment (TorstenDittmann) [inline src/routes/console/project-[project]/overview/+layout.svelte:92]: the one for documents should be fine 👍🏻 they dont need any change

Comment (TorstenDittmann) [reply src/routes/console/project-[project]/overview/+layout.svelte:92]: I think only executions are different

TorstenDittmann (changes requested):
👍🏻

src/routes/console/project-[project]/overview/+layout.svelte
Comment (lohanidamodar) [reply src/routes/console/project-[project]/overview/+layout.svelte:92]: All these should be summed, as all these are now using time-series data, before we got sum from database and data already had the sum, now it's not

lohanidamodar (approved):

src/routes/console/project-[project]/overview/+layout.svelte
Comment (Meldiron) [reply src/routes/console/project-[project]/overview/+layout.svelte:92]: I believe that after usage refactor the is no more aggregation and we ONLY have time series data. So we should be summing everything.

Comment (Meldiron) [reply src/routes/console/project-[project]/overview/+layout.svelte:92]: Of course, not the best solution since we don't have real total, just total in time period. But its a patch until usage container can properly sum.

christyjacob4 (approved):

PR #206, opened 2022-12-05
Fix: copy not working on http

TGlide:
## What does this PR do?

This PR fixes a bug in which the copy component would trigger an error on non-local HTTP endpoints. To fix this, a unsecuredCopy method is introduced, which is called when the previous copy implementation doesn't work.

## Test Plan

Manual

## Related PRs and Issues

N/A

Yes

### DIFF
--- a/src/lib/components/copy.svelte
+++ b/src/lib/components/copy.svelte
@@ -2,6 +2,7 @@
     import { trackEvent } from '$lib/actions/analytics';
     import { tooltip } from '$lib/actions/tooltip';
     import { clickOnEnter } from '$lib/helpers/a11y';
+    import { copy } from '$lib/helpers/copy';
     import { addNotification } from '$lib/stores/notifications';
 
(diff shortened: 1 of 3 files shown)


src/lib/components/copy.svelte
```diff
@@ -9,23 +9,54 @@
...
+            success = false;
+        } finally {
+            document.body.removeChild(textArea);
+        }
+
+        return success;
+    }
+
+    async function copy() {
+        // Because of how JS works, unsecuredCopy only runs if securedCopy fails
```
Meldiron, line 42:
```suggestion
        // securedCopy works only in HTTPS environment. unsecuredCopy is used as backup for HTTP.
```

Meldiron (approved):

ArmanNik (approved):

TorstenDittmann (approved):

PR #216, opened 2022-12-12
Feat: empty arrays UI

TGlide:
## What does this PR do?

Creates an AttributeForm component, that substitutes the forms on create document and update document. This component also changes the logic of array fields, allowing documents with empty array attributes to be created.

## Test Plan

Manual

## Related PRs and Issues

N/A

Yes


TorstenDittmann (changes requested):
Bildschirm­foto 2023-01-10 um 15 15 23

the *Add item* button is not vertically aligned with the attribute name 👍🏻

TGlide:
> Bildschirm­foto 2023-01-10 um 15 15 23
> 
> the _Add item_ button is not vertically aligned with the attribute name 👍🏻

I can't seem to reproduce:
image


Any particular steps you went through?

src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/store.ts
```diff
@@ -12,6 +13,12 @@ export type Attributes =
     | Models.AttributeString
     | Models.AttributeUrl;
 
+export function isAttributeEnum(attribute: unknown): attribute is Models.AttributeEnum {
+    return isObjectType<Models.AttributeEnum>(attribute, {
+        format: (v) => v === 'enum'
+    });
+}
+
```
Comment (TorstenDittmann) [inline src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/store.ts:21]: I am not sure we need this here. because this would also apply for email attribute as well. I think a simple check for `format` should be enough

Comment (TGlide) [reply src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/store.ts:21]: Do you mean we don't need the type guard, or that we don't need the isObjectType, and instead should only check if format exists?
If it's the latter, than the type check would be kinda false, because as you said, it could apply to Models.AttributeEmail

src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/[[page]]/+page.svelte
```diff
@@ -33,6 +34,44 @@
             title: attribute.key
         }))
     ];
+
+    function formatArray(array: any[]) {
```
Comment (TorstenDittmann) [inline src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/[[page]]/+page.svelte:38]:
```suggestion
    function formatArray(array: unknown[]) {
```

src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/[[page]]/+page.svelte
```diff
@@ -33,6 +34,44 @@
...
+                formattedFields.push(`"${item}"`);
+            } else {
+                formattedFields.push(`${item}`);
+            }
+        }
+
+        return `[${formattedFields.join(', ')}]`;
+    }
+
+    function formatColumn(column: any) {
```
Comment (TorstenDittmann) [inline src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/[[page]]/+page.svelte:53]:
```suggestion
    function formatColumn(column: unknown) {
```

TorstenDittmann (approved):

PR #217, opened 2022-12-12
feat: display attribute types on input fields

TGlide:
## What does this PR do?

- Displays attribute types on input fields
- Also includes some changes to the documents table, after talking with design

## Test Plan

Manual, plus tests for two new helper functions

## Related PRs and Issues

- #216 

Yes

### DIFF
--- a/src/lib/helpers/string.ts
+++ b/src/lib/helpers/string.ts
@@ -1,0 +1,10 @@
+/**
+ * Capitalizes the first letter of a string
+ *
+ * @export
+ * @param {string} str
+ * @returns {string}
+ */
+export function capitalize(str: string): string {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+}
--- a/src/lib/actions/tooltip.ts
+++ b/src/lib/actions/tooltip.ts
@@ -1,17 +1,24 @@
 import type { Action } from 'svelte/action';
-import type { Props } from 'tippy.js';
+import type { Props as TippyProps } from 'tippy.js';
 import tippy from 'tippy.js';
+
+type Props = TippyProps & {
+    disabled?: boolean;
+};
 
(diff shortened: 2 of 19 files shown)


src/lib/helpers/string.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Capitalizes the first letter of a string
+ *
+ * @export
+ * @param {string} str
+ * @returns {string}
+ */
+export function capitalize(str: string): string {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+}
```
TorstenDittmann, line 10: It might be smarter to use CSS for this 🤔

TGlide: The behaviour is different. With CSS, the first letter of every word is capitalized. This function onnly does it for the first word. The question is, which one is preferred?

TorstenDittmann: iirc there should be a `:first-letter` selector 👀

TGlide: Oh, TIL! Will use CSS then, for sure :smile:

TGlide: @TorstenDittmann After further inspection, the way I see it, in order to use CSS I'd have to either pass in a custom prop to the inputs being used to Attribute, or use :global() in AttributeForm's styles. With this in mind, I think we're better off keeping this as is. What do you think?

TorstenDittmann:
Bildschirm­foto 2023-01-27 um 13 30 39

Code wise looks good! But there is a small issue with the event text input 👀

src/lib/helpers/string.ts
TorstenDittmann: nah, wanted to make it simpler - this is not the way :-D

TorstenDittmann (approved):

PR #246, opened 2023-02-02
Feat: function execution timer

TGlide:
## What does this PR do?

Adds a timer to in-progress function executions

## Test Plan

Manual

## Related PRs and Issues

N/A

Yes


tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
     "extends": "./.svelte-kit/tsconfig.json",
     "compilerOptions": {
-        "lib": ["webworker"],
+        "lib": [
+            "webworker",
+            "DOM",
+            "ESNext"
```
TorstenDittmann, line 7: Do we even need those? I am pretty sure they are present in `"extends": "./.svelte-kit/tsconfig.json"`

TorstenDittmann (commented):
There are some linting issues 👍🏻

tsconfig.json
TGlide: They are, but by defining lib as webworker only, we're overriding svelte-kit's tsconfig's lib option, so we need to re-include it.

TorstenDittmann: ah okay, thought its actually extending :-D

TorstenDittmann (approved):

PR #263, opened 2023-02-08
feat: add even modal description

ArmanNik:


src/lib/components/eventModal.svelte
```diff
@@ -240,7 +240,14 @@
...
     <svelte:fragment slot="header">Create Event</svelte:fragment>
-
+    <p class="text">
+        Select events in your Appwrite project that will trigger your function. <a
+            href="https://appwrite.io/docs/events"
+            target="_blank"
+            rel="noopener noreferrer"
+            class="link">Learn more about Appwrite Events</a
+        >.
+    </p>
```
TorstenDittmann, line 250: That is now also shown for webhooks 😉

PR #287, opened 2023-02-12
fix: enum attribute editable issue in overview modal

yatharth1706:
## What does this PR do?

It fixes the issue where enum attributes were editable in overview modal. 

## Test Plan

Tested manually by creating enum attribute in localhost and tried editing the enum input tag by writing or removing the tags.

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/5009

### Have you read the [Contributing Guidelines on issues

Yes, I have read the contributing guidelines

### DIFF
--- a/src/lib/elements/forms/inputTags.svelte
+++ b/src/lib/elements/forms/inputTags.svelte
@@ -89,7 +89,7 @@
                                     type="button"
                                     class="input-tag-delete-button"
                                     aria-label={`delete ${tag} tag`}
-                                    on:click={() => removeValue(tag)}>
+                                    on:click={() => !readonly && removeValue(tag)}>
                                     <span class="icon-x" aria-hidden="true" />
                                 </button>
                             </div>


src/lib/elements/forms/inputTags.svelte
```diff
@@ -89,7 +89,7 @@
                                     type="button"
                                     class="input-tag-delete-button"
                                     aria-label={`delete ${tag} tag`}
-                                    on:click={() => removeValue(tag)}>
+                                    on:click={() => !readonly && removeValue(tag)}>
```
TorstenDittmann, line 92: I think the `readonly` check should be in the `removeValue` function. Otherwise you can remove tags with the backspace key 👍🏻 

https://github.com/appwrite/console/blob/24b8e2221107aac2db73a7ac1fb2c10607d90488/src/lib/elements/forms/inputTags.svelte#L53-L55

```
if (readonly) return;
```
should to the trick 👍🏻

TorstenDittmann (changes requested):

src/lib/elements/forms/inputTags.svelte
yatharth1706: @TorstenDittmann thank you for the suggestion. Have done the change in code

TorstenDittmann (approved):

PR #317, opened 2023-02-21
chore: update dependencies

ArmanNik:


src/lib/elements/forms/button.svelte
```diff
@@ -40,11 +37,11 @@
 {#if href}
     <a
         on:click={track}
-        {...attributes}
         {href}
         target={external ? '_blank' : ''}
         rel={external ? 'noopener noreferrer' : ''}
         class="button"
+        class:is-disabled={disabled}
```
TorstenDittmann, line 44:
```suggestion
        class:is-disabled={internalDisabled}
```

ArmanNik: @TorstenDittmann Not sure this change is needed... if for example, we have links inside of modals they would be disabled needlessly disabled :thinking:

TorstenDittmann (approved):

PR #332, opened 2023-03-09
Refactor: add truncate option to output

TGlide:
## What does this PR do?

Adds truncate option to output

## Test Plan

Manual

## Related PRs and Issues

- https://github.com/appwrite/pink/pull/63

Yes


TGlide:
> Shouldn't we truncate according to the container size instead of the number of characters? 🤔

You're right, maybe I could use an Action to do it... 👀

PR #339, opened 2023-03-14
Feat: notification transitions

TGlide:
## What does this PR do?

Adds transitions to notifications

https://user-images.githubusercontent.com/26071571/225066061-e89fecb7-7796-4806-8c32-2cbf75e3d9cf.mp4

## Test Plan

Manual

## Related PRs and Issues

N/A

Yes


ArmanNik (approved):

heyCarla:
The animation looks great! 👨‍🍳👌 
Is there a reason why we show only three at a time? Or was it just the timing for the dismissal? If feasible, I think showing all alerts when multiple get triggered simultaneously would be more helpful. Otherwise the user might miss something important about future actions, errors, etc.

ArmanNik:
> The animation looks great! man_cookok_hand Is there a reason why we show only three at a time? Or was it just the timing for the dismissal? If feasible, I think showing all alerts when multiple get triggered simultaneously would be more helpful. Otherwise the user might miss something important about future actions, errors, etc.

@heyCarla if I remember correctly, the max number was decided by the design team :upside_down_face:

heyCarla (approved):
Am fine with the transitions, but let's discuss the number shown in our next FE sync 😄

PR #349, opened 2023-03-17
Feat: input count dynamic color

TGlide:
## What does this PR do?

- Changes input color based on value
- Use textareas instead of inputs for text inputs with size > 64

## Test Plan

Manual

## Related PRs and Issues

N/A

Yes

### DIFF
--- a/src/lib/elements/forms/inputTextarea.svelte
+++ b/src/lib/elements/forms/inputTextarea.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
     import { onMount } from 'svelte';
     import { FormItem, Helper, Label } from '.';
+    import TextCounter from './textCounter.svelte';
 
     export let label: string;
     export let showLabel = true;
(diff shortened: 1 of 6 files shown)


src/lib/elements/forms/inputTextarea.svelte
```diff
@@ -53,8 +55,33 @@
...
 </FormItem>
+
+<style class="scss">
+    .text-counter {
+        --p-text-counter-color: var(--color-neutral-50);
+    }
+
+    .text-counter[data-active='true'] {
+        --p-text-counter-color: var(--color-neutral-70);
+    }
```
TorstenDittmann, line 78: wouldn't be adding a class for this more elegant?

TGlide: Do you mean adding a class in Pink? If not, IMO data-attributes make more sense than classes for controlling element 'states', but it's personal preference only, I can add a class if you prefer.

TorstenDittmann: we can just add classes in the svelte component here 👍🏻

TorstenDittmann (approved):

PR #369, opened 2023-03-28
fix: limit sidenav transition

TGlide:
## What does this PR do?

- Fixes the secondary sidenav having an animation when there's no parent
- Keeps sidenav open when returning from sceondSideNav

## Test Plan

Manual

## Related PRs and Issues

N/A

Yes

### DIFF
--- a/src/lib/layout/shell.svelte
+++ b/src/lib/layout/shell.svelte
@@ -1,17 +1,17 @@
 <script lang="ts">
-    import { navigating, page } from '$app/stores';
     import { browser } from '$app/environment';
-    import { wizard } from '$lib/stores/wizard';
-    import { log } from '$lib/stores/logs';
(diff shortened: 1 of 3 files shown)


src/lib/layout/shell.svelte
```diff
@@ -1,17 +1,17 @@
...
     export let isOpen = false;
     export let showSideNavigation = false;
 
     let y: number;
 
-    navigating.subscribe(() => {
-        if (isOpen) isOpen = false;
+    page.subscribe(() => {
+        isOpen = $page.url.searchParams.get('openNavbar') === 'true';
     });
```
TorstenDittmann, line 15: very nitpicky, but if we subscribe to that store - lets actually take its value in the callback instead of `$page`

TorstenDittmann (approved):

PR #377, opened 2023-04-02
Fix 5295 code component style fix

kabhamo:
## What does this PR do?

The fix is simple, adjust the `padding-top` style, so the **clipboard icon** has more space.
I have ensured that the **code space** and all the elements have flexible sizes relevant to the screen size.
see the picture below
 
Screenshot 2023-03-30 at 2 07 01

Screenshot 2023-03-30 at 2 06 50

## Test Plan

Run the console dashboard on the localhost 
create new project
create new platform
click next till you reach the "Let's coding" step

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/5295

Yes


ArmanNik:
@kabhamo instead of using `padding-top` let's use `padding-block-start` :) 
Also, simply removing the padding property has pretty much the same effect :thinking: @chenparnasa was the tag overlap ping the code intended?

kabhamo:
Because when decreasing the screen size it will crash the style.
However I will test the css property u mentioned ASAP.

kabhamo:
@ArmanNik, I have changed the property to `padding-block-start`.
It works as needed, please review the code and tell me if anything needed.

chenparnasa:
@ArmanNik The behavior now is similar to the old code snippets we are using (in the docs for example). We might iterate this component in the future, but for now, we can increase the padding so the code and tag won't overlap

TorstenDittmann (commented):
can you fix the linter issues with `npm run format` 🙏🏻

kabhamo:
@stnguyen90 yeah sure i will do it

PR #406, opened 2023-04-16
Reset custom validity on re-render if required checkbox field unchecked again.

Akshay-Rana-Gujjar:
While creating attribute in a database using console ui, if user click on required checkbox and if uncheck it again then default value do not accept any value from the default value dropdown.

## What does this PR do?

This PR will fix the issue when user uncheck the required checkbox field then user is not able to select default value of attribute because we are not resetting custom validity in InputSelect element in form.

## Test Plan

https://user-images.githubusercontent.com/24373771/232279103-47a4c997-eaba-450c-a7e7-e9db4630a293.mov

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/5366

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?
Yes

PR #407, opened 2023-04-17
feat: update select validation for boolean attributes

christyjacob4:
## What does this PR do?

The validation for the `select` input did not take into account that `false` is a valid value for boolean attribiutes

### Before
Screenshot

### After
Screenshot 2023-04-17 at 11 48 50 AM

## Test Plan

## Related PRs and Issues

https://github.com/appwrite/appwrite/issues/5389

(Write your answer here.)

### DIFF
--- a/src/lib/elements/forms/inputSelect.svelte
+++ b/src/lib/elements/forms/inputSelect.svelte
@@ -27,29 +27,17 @@
         error = element.validationMessage;
     };
 
-    const isValid = (value: string | number | boolean) => {
-        if (value === null) return false;
-
-        if (value === undefined) return false;
-
-        if (typeof value !== 'boolean') {
-            if (!value) {
-                return false;
-            }
-        }
-
-        return true;
+    const isNotEmpty = (value: string | number | boolean) => {
+        return typeof value === 'boolean' ? true : !!value;
     };
 
-    $: if (element && required && !isValid(value)) {
-        element.setCustomValidity('This field is required');
-    }
-
-    $: if (element && required && isValid(value)) {
-        element.setCustomValidity('');
-    }
-
-    $: if (isValid(value)) {
+    $: if (element && required) {
+        if (!isNotEmpty(value)) {
+            element.setCustomValidity('This field is required');
+        } else {
+            element.setCustomValidity('');
+        }
+    } else if (isNotEmpty(value)) {
         error = null;
     }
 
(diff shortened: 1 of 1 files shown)


src/lib/elements/forms/inputSelect.svelte
```diff
@@ -27,15 +27,29 @@
...
+        if (typeof value !== 'boolean') {
+            if (!value) {
+                return false;
+            }
+        }
+
+        return true;
+    };
+
+    $: if (element && required && !isValid(value)) {
```
TGlide, line 44: This can be better expressed, see https://github.com/appwrite/console/pull/406#issuecomment-1511673923

src/lib/elements/forms/inputSelect.svelte
```diff
@@ -27,15 +27,29 @@
...
+        if (value === undefined) return false;
+
+        if (typeof value !== 'boolean') {
+            if (!value) {
+                return false;
+            }
+        }
+
+        return true;
+    };
```
TGlide, line 42:
```suggestion
    const isNotEmpty = (value: string | number | boolean) => {
        return typeof value === 'boolean' ? true : !!value
    };
```
I think this works, no?

TGlide (changes requested):

src/lib/elements/forms/inputSelect.svelte
christyjacob4: Do you want it to always return `true` if the type is `boolean`? Would there be a case where a boolean value is `null` or `undefined`?

TGlide: `typeof null` returns `object` and `typeof undefined` returns `undefined`, so `typeof value === 'boolean'` only is true when value is either `true` or `false`

src/lib/elements/forms/inputSelect.svelte
TGlide: Yes, 100%! I just meant that we can use a single `$` sign with `if...else if... else`.

christyjacob4: updated 👌  Lemme know if this is what you meant

TGlide: It's similar:
```ts
$: if (required && !isNotEmpty(value)) {
    element?.setCustomValidity("This field is required");
  } else {
    element?.setCustomValidity("");
  }

  $: if (isNotEmpty(value)) {
    error = null;
  }
```
With the ?. after element, we don't need to check if element exists.
I kept the `error = null` to separate concerns

christyjacob4: updated 👌  Thanks for the suggestion :)

TGlide (approved):

TorstenDittmann (approved):

PR #427, opened 2023-05-02
feat-cloud-gtm

ArmanNik:


vite.config.ts.timestamp-1682949685828.mjs
TorstenDittmann, line 1: whats that file for? 👀

ArmanNik: Idk... it was there so I left it :rofl:

TorstenDittmann (approved):

PR #447, opened 2023-05-18
Clarify Deleting Account

stnguyen90:
## What does this PR do?

Console accounts are not deleted, so a user won't be able to re-create their account after "deleting" it.

## Test Plan

Manual:

image

image

image

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/5490 

Yes

### DIFF
--- a/src/routes/console/account/+page.svelte
+++ b/src/routes/console/account/+page.svelte
@@ -154,10 +154,10 @@
     </Form>
     <CardGrid danger>
         <div>
-            <Heading tag="h6" size="7">Delete Account</Heading>
+            <Heading tag="h6" size="7">Deactivate Account</Heading>
         </div>
         <p>
-            Your account will be permanently deleted and access will be lost to any of your teams
+            Your account will be permanently deactivated and access will be lost to any of your teams
             and data. This action is irreversible.
         </p>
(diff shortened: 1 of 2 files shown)


ArmanNik (changes requested):
@stnguyen90 let's also clear all the locally stored data (both local and session storage), otherwise the user won't be able to login with another account

I would also suggest adding a checkbox for confirmation the same way we do it for documents with relationships :smile:

stnguyen90:
@ArmanNik 

> let's also clear all the locally stored data (both local and session storage), otherwise the user won't be able to login with another account

Yep! I plan on submitting a separate PR for that today.

> I would also suggest adding a checkbox for confirmation the same way we do it for documents with relationships

I want to make sure we can roll this out to cloud...are there styling issues with checkboxes for the version cloud is on?

image

ArmanNik:
There are still issues with the checkbox style on Cloud... the solution is to update Cloud to the latest version of Appwrite 🙃

ArmanNik (approved):

eldadfux:
Not sure about this feature, let's wait. It might be better to just allow deletion and trust backups for any abusive play.

PR #475, opened 2023-07-07
Show user's last activity

stnguyen90:
## What does this PR do?

Show users' last activity on users list and detail pages.

## Test Plan

Manual:

### User Last Accessed

image

image

image

### API Key Last Accessed

image

image

image

## Related PRs and Issues

Depends on new Console SDK release and

* https://github.com/appwrite/appwrite/pull/5781

Yes

### DIFF
--- a/src/lib/helpers/date.ts
+++ b/src/lib/helpers/date.ts
@@ -1,5 +1,10 @@
 export const toLocaleDate = (datetime: string) => {
     const date = new Date(datetime);
+
+    if (isNaN(date.getTime())) {
+        return 'n/a';
+    }
+
     const options: Intl.DateTimeFormatOptions = {
         year: 'numeric',
         month: 'short',
@@ -11,6 +16,11 @@ export const toLocaleDate = (datetime: string) => {
 
 export const toLocaleDateTime = (datetime: string | number) => {
     const date = new Date(datetime);
+
+    if (isNaN(date.getTime())) {
+        return 'n/a';
+    }
+
     const options: Intl.DateTimeFormatOptions = {
         year: 'numeric',
         month: 'short',
--- a/src/routes/console/project-[project]/auth/+page.svelte
+++ b/src/routes/console/project-[project]/auth/+page.svelte
@@ -19,7 +19,7 @@
         TableRowLink
     } from '$lib/elements/table';
     import { Pill } from '$lib/elements';
-    import { toLocaleDateTime } from '$lib/helpers/date';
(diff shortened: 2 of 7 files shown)


eldadfux:
@stnguyen90 we should omit the time, the user update method happens only once a day and we record the time of the first activity of the day, starting with the 2nd activity the time is out of sync and misleading. The next update will only happen the next day.

stnguyen90:
> @stnguyen90 we should omit the time, the user update method happens only once a day and we record the time of the first activity of the day, starting with the 2nd activity the time is out of sync and misleading. The next update will only happen the next day.

@eldadfux, that's a good idea to reduce confusion. FYI, we still might have some confusing data due to timezones.

eldadfux:
Why can't we take care of timezone on the console side? we should have the information on the timestamp returned from the API. Also, do we explain in the API docs that this timestamp updates once a day?

Anyway, I strongly believe we need to omit the time, the first question we'll get on Discord is why the user activity isn't updating when they see the time is stuck throughout the day. 

BTW, we also have the same issue with API key, we need to fix both.

stnguyen90:
@eldadfux, 

> Why can't we take care of timezone on the console side? we should have the information on the timestamp returned from the API.

Let's say there's been recent activity, but the timestamp is today at `2023-07-20T02:00:00.000Z` because it's still within 24 hours. Since I'm `UTC-07:00`, when the timestamp is localized, it becomes 2023-07-19 (yesterday). 

> Also, do we explain in the API docs that this timestamp updates once a day?

Here's a PR: https://github.com/appwrite/appwrite/pull/5855.

eldadfux:
> @eldadfux,
> 
> > Why can't we take care of timezone on the console side? we should have the information on the timestamp returned from the API.
> 
> Let's say there's been recent activity, but the timestamp is today at `2023-07-20T02:00:00.000Z` because it's still within 24 hours. Since I'm `UTC-07:00`, when the timestamp is localized, it becomes 2023-07-19 (yesterday).
> 
> > Also, do we explain in the API docs that this timestamp updates once a day?
> 
> Here's a PR: [appwrite/appwrite#5855](https://github.com/appwrite/appwrite/pull/5855).

This is fine for me, the goal here was just to show a rough range of your activity. We didn't plan for it to be accurate to the minute. We need to do the same for the API key `LAST ACCESSED` column.

stnguyen90:
> We need to do the same for the API key LAST ACCESSED column.

@eldadfux, I updated it for API keys too.

eldadfux (approved):

TorstenDittmann (approved):

PR #495, opened 2023-08-08
fixed issue that didn't allow a user to update a document when deleting elements in an array attribute

safwanyp:
Use `symmetricDifference` instead of `difference`.
- `difference` only returns values that are present in the edited attributes, but not in the originally fetched attribute vales. This always ends up returning an empty array.
- I used `symmetricDifference` as it returns an intersection of `difference` from both sides and works perfectly for this use case, as the original attributes never change.

## What does this PR do?

This PR closes https://github.com/appwrite/appwrite/issues/5476

## Test Plan

- In addition to the existing tests, I performed manual testing by adding and deleting elements in an array attribute. 
- I was going to change the way the `difference` function worked, but realized it was being used in other places, which broke things. I used the `symmetricDifference` function as it fits in perfectly for this use case.

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/5476

Yes ✅

### DIFF
--- a/src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/document-[document]/data/+page.svelte
+++ b/src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/document-[document]/data/+page.svelte
@@ -15,6 +15,7 @@
     import AttributeItem from '../attributeItem.svelte';
     import { symmetricDifference } from '$lib/helpers/array';
     import { isRelationship, isRelationshipToMany } from '../attributes/store';
+    import { deepClone } from '$lib/helpers/object';
 
     const databaseId = $page.params.database;
     const collectionId = $page.params.collection;
@@ -22,21 +23,23 @@
     const editing = true;
 
     const work = writable(
-        Object.keys($doc)
-            .filter((key) => {
-                return ![
-                    '$id',
-                    '$collection',
-                    '$collectionId',
-                    '$databaseId',
-                    '$createdAt',
-                    '$updatedAt'
-                ].includes(key);
-            })
-            .reduce((obj, key) => {
-                obj[key] = $doc[key];
-                return obj;
-            }, {}) as Models.Document
+        deepClone(
+            Object.keys($doc)
+                .filter((key) => {
+                    return ![
+                        '$id',
+                        '$collection',
+                        '$collectionId',
+                        '$databaseId',
+                        '$createdAt',
+                        '$updatedAt'
+                    ].includes(key);
+                })
+                .reduce((obj, key) => {
+                    obj[key] = $doc[key];
+                    return obj;
+                }, {}) as Models.Document
+        )
     );
 
     async function updateData() {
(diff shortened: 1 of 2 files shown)


safwanyp:
Okay I just realized there's another bug. 
Although elements can be deleted, existing items cannot be edited. The `Update` button stays disabled even after changing values. Should I create a new issue for that, or push a commit to fix it in this PR?

safwanyp:
Okay so I figured out the issue that prevents updating of a document when editing the elements of an array attribute. 

In the code it seems like `$work` stores the value of edited values. `$work` is populated by iterating over `$doc`, but they are referencing the same underlying value, no? Due to this, when `$work` is populated, `$doc` is updated as well, which results in both the arrays having no difference. 

I can think of 2 solutions to this: 
1. Use `JSON.parse(JSON.stringify($doc))` to clone the $doc, and not reference copy.
2. Use a locally decalred variable instead of `formValues[attribute.key][index]` in `<AttributeItem />`, which can be passed into the `<Attribute />` component. Obviously, the local variable will be updated when there's a change.

I have only tested with the first approach as it doesn't mess with the functionality outside of the required component.

TGlide:
> Okay so I figured out the issue that prevents updating of a document when editing the elements of an array attribute.
> 
> In the code it seems like `$work` stores the value of edited values. `$work` is populated by iterating over `$doc`, but they are referencing the same underlying value, no? Due to this, when `$work` is populated, `$doc` is updated as well, which results in both the arrays having no difference.
> 
> I can think of 2 solutions to this:
> 
>     1. Use `JSON.parse(JSON.stringify($doc))` to clone the $doc, and not reference copy.
> 
>     2. Use a locally decalred variable instead of `formValues[attribute.key][index]` in `<AttributeItem />`, which can be passed into the `<Attribute />` component. Obviously, the local variable will be updated when there's a change.
> 
> 
> I have only tested with the first approach as it doesn't mess with the functionality outside of the required component.

Hi @safwanyp! Thank you for the PR 😄 

I'd say doing a deepClone with JSON.parse should be enough. I'd recommend creating an utility for it in `helpers/object.ts`

safwanyp:
Hey @TGlide, thanks for the feedback and pointers!

I've just pushed a commit that creates a `deepClone` utility function, and uses it to create a copy of `$doc` into `$work`. Tests were successful on local copy. Appreciate any feedback!

src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/document-[document]/data/+page.svelte
```diff
@@ -13,30 +13,33 @@
...
+                        '$databaseId',
+                        '$createdAt',
+                        '$updatedAt'
+                    ].includes(key);
+                })
+                .reduce((obj, key) => {
+                    obj[key] = $doc[key];
+                    return obj;
+                }, {}) as Models.Document
+        )
```
Comment (TGlide) [inline src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/document-[document]/data/+page.svelte:42]: I think this function has become quite verbose.

As an alternative, we can create an initWork function, where we have better control flow.

e.g.

```ts
function initWork() {
  let prohibitedKeys = ['$id', ...];
  const filteredKeys = Object.keys($doc)
  // So on and so forth..
}

const work = initWork()
```

Comment (safwanyp) [reply src/routes/console/project-[project]/databases/database-[database]/collection-[collection]/document-[document]/data/+page.svelte:42]: that's fair. I'll work on this 👍🏻

TGlide (approved):

christyjacob4 (approved):

PR #479, opened 2023-07-17
Command center and migrations

TGlide:
## What does this PR do?

- Adds Command center
- Adds migration support

PR #498, opened 2023-08-09
add title and description props and slots for Modal

TGlide:
## What does this PR do?

Refactors Modal to allow title and description in header, just like in [Pink](https://pink.appwrite.io/components/modal).

I also opted to allow a prop instead of a slot, while keeping the slot as an option. The prop is autocompletable, unlike the slot, and is much simpler to write than `<svelte:fragment slot="title>`. The slot should remain though IMo in cases where you want to add HTML elements rather than just text.

This refactors most Modals to use the prop syntax.

## Test Plan

Manual

## Related PRs and Issues

Based on #479

Yes


ArmanNik (changes requested):
We need to refactor all remaining modals introduced after this PR :sweat_smile:

ArmanNik (approved):

PR #522, opened 2023-08-29
fix: qa last fixes

ArmanNik:
Fix variable creation during function creation. 
Feat dynamic component for variables

### DIFF
--- a/src/lib/wizards/functions/createTemplate.svelte
+++ b/src/lib/wizards/functions/createTemplate.svelte
@@ -32,6 +32,7 @@
                 );
                 $templateConfig.variables['APPWRITE_API_KEY'] = key.secret;
             }
+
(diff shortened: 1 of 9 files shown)


src/lib/wizards/functions/createTemplate.svelte
```diff
@@ -54,6 +55,17 @@
...
             );
+
+            if ($templateConfig.variables) {
+                for (const [key, value] of Object.entries($templateConfig.variables)) {
+                    await sdk.forProject.functions.createVariable(
+                        response.$id,
+                        key,
+                        value?.toString()
+                    );
+                }
```
TorstenDittmann, line 66: can we use a `Promise.all` helper for better performance?

TorstenDittmann (approved):

PR #542, opened 2023-09-12
fix: function templates runtimes

ArmanNik:
## What does this PR do?

Fix function templates runtimes

## Test Plan

Manual

Yes

### DIFF
--- a/src/lib/wizards/functions/cover.svelte
+++ b/src/lib/wizards/functions/cover.svelte
@@ -7,16 +7,12 @@
             variables[variable.name] = variable.value ?? '';
         });
 
-        if (!language) {
-            language = template.runtimes[0].name;
-        }
-
         templateStore.set(template);
         templateConfig.set({
             $id: null,
+            name: template.name,
             language,
             runtime: null,
-            name: template.name,
             variables,
             repositoryBehaviour: 'new',
             repositoryName: template.id,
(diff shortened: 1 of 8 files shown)


src/lib/wizards/functions/cover.svelte
```diff
@@ -106,12 +106,15 @@
...
                         {:then response}
-                            {@const runtimes = new Map(response.runtimes.map((r) => [r.$id, r]))}
-                            {@const templates = quickStart.runtimes.filter((_template) =>
-                                runtimes.has(_template.name)
-                            )}
+                            {@const templates = quickStart.runtimes.filter((_template) => {
+                                return response.runtimes.some((runtime) =>
+                                    runtime.$id.includes(_template.name)
+                                );
+                            })}
```
stnguyen90, line 113: This looks like it reverted a previous change. We should make sure to use better data types when available to avoid doing n^2 operations (iterating over an array in an array).

ArmanNik: But now the template name is only going to include the language used not the version so `has` will not work anymore :thinking:

Meldiron:
I think we can close this PR in favour of: https://github.com/appwrite/console/pull/546

We came to conclusion we need to keep it version-aware.

stnguyen90:
I've split the SMTP change out of this PR and moved it over https://github.com/appwrite/console/pull/548

PR #544, opened 2023-09-13
Update create function cover for case where VCS is not enabled

stnguyen90:
## What does this PR do?

1. Simplify usage of console variables
2. Update create function cover for case where VCS is not enabled

Closes https://github.com/appwrite/appwrite/issues/6180

## Test Plan

Manual w/o VCS:

vcsDisabled Create Function

Manual w/ VCS:

vcsEnabled Create Function

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/6180

Yes

### DIFF
--- a/src/lib/pages/domains/wizard/step1.svelte
+++ b/src/lib/pages/domains/wizard/step1.svelte
@@ -6,21 +6,12 @@
     import { sdk } from '$lib/stores/sdk';
     import { isSelfHosted } from '$lib/system';
     import { func } from '$routes/console/project-[project]/functions/function-[function]/store';
-    import { onMount } from 'svelte';
     import { ProxyTypes } from '../index.svelte';
     import { domain, typeStore } from './store';
     import { consoleVariables } from '$routes/console/store';
 
     let error = null;
-    let isDomainsEnabled = false;
-
-    onMount(async () => {
-        if (!isSelfHosted) {
-            return;
-        }
-
-        isDomainsEnabled = (await $consoleVariables)?._APP_DOMAIN_ENABLED === true;
-    });
+    const isDomainsEnabled = $consoleVariables?._APP_DOMAIN_ENABLED === true;
 
     async function createDomain() {
         try {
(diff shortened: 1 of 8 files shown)


src/lib/pages/domains/wizard/step1.svelte
```diff
@@ -6,21 +6,12 @@
...
-    let isDomainsEnabled = false;
-
-    onMount(async () => {
-        if (!isSelfHosted) {
-            return;
-        }
-
-        isDomainsEnabled = (await $consoleVariables)?._APP_DOMAIN_ENABLED === true;
-    });
+    const isDomainsEnabled = $consoleVariables?._APP_DOMAIN_ENABLED === true;
```
ArmanNik, line 14: `$consoleVariables` might need to be awaited since we are not doing it in the layout

ArmanNik (changes requested):
I don't think it makes sense to await the consoleVariables in the main layout.ts we can await them when we need them :+1:

src/routes/console/+layout.ts
```diff
@@ -16,7 +16,7 @@ export const load: LayoutLoad = async ({ fetch, depends }) => {
     const data = await response.json();
 
     return {
-        consoleVariables: sdk.forConsole.console.variables(),
+        consoleVariables: await sdk.forConsole.console.variables(),
```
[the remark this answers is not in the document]
stnguyen90: I'm not sure if it's worth it because:

1. it complicates the rest of the code
2. the variables are fetched 20ms after version even before this change

What could be useful is fetching both the version and the variables concurrently.

src/lib/wizards/functions/cover.svelte
```diff
@@ -81,146 +130,164 @@
...
+                                        </button>
+                                    </li>
+                                {/each}
+                            {:then response}
+                                {@const runtimes = new Map(
+                                    response.runtimes.map((r) => [r.$id, r])
+                                )}
+                                {@const templates = quickStart.runtimes.filter((_template) =>
+                                    runtimes.has(_template.name)
+                                )}
```
TorstenDittmann, line 162:
```suggestion
                                {@const templates = quickStart.runtimes.filter((template) =>
                                    runtimes.has(template.name)
                                )}
```

stnguyen90: updated

ArmanNik (approved):

PR #565, opened 2023-09-24
Fix-248-Copy-button-issue

btme0011:
## What does this PR do?

Shows Copy pop while hovering on the copy button and shows 'copied' when button is clicked

## Test Plan

Tested locally 

Screenshot 2023-09-24 at 10 49 08 PM
Screenshot 2023-09-24 at 10 49 01 PM

## Related PRs and Issues

 #6248

(Write your answer here.)

### DIFF
--- a/src/lib/components/copyInput.svelte
+++ b/src/lib/components/copyInput.svelte
@@ -24,7 +24,7 @@
     };
 </script>
 
-<div>
+<div id="tooltip-container">
     <label class:u-hide={!showLabel} class="label" for={label}>{label}</label>
     <div class="input-text-wrapper" style="--amount-of-buttons:1">
         <input {value} id={label} type="text" class="input-text" readonly />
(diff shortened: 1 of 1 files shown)


src/lib/components/copyInput.svelte
```diff
@@ -24,7 +24,7 @@
     };
 </script>
 
-<div>
+<div id="tooltip-container">
```
stnguyen90, line 27: It's best to avoid IDs when possible since they're highly selective regarding CSS, and IDs should be unique, but this component can be used multiple times/places. For example,

https://github.com/appwrite/console/blob/ffb7476b28e55b8fbffb57626bfa4f8cdbdaf821/src/routes/console/project-%5Bproject%5D/settings/updateName.svelte#L48-L49

src/lib/components/copyInput.svelte
```diff
@@ -37,7 +37,8 @@
                 on:mouseenter={() => setTimeout(() => (content = 'Click to copy'))}
                 use:tooltip={{
                     content,
-                    hideOnClick: false
+                    hideOnClick: false,
+                    appendTo: document.getElementById('tooltip-container')
```
stnguyen90, line 41: Nice find! I do think this is on the right track. Looking at the documentation, appendTo can be a function that takes a reference to the element itself, so we can change this to:

```suggestion
                    appendTo: (ref) => ref.parentElement.parentElement
```

Be sure to add a comment or commit message to explain why we need to do `ref.parentElement.parentElement`..

stnguyen90 (changes requested):
Great PR! 🤯 We left some comments during the review, please check them out.

PR #571, opened 2023-10-01
Fix wrong validation of projectid

nick2432:
## What does this PR do?

update the helper text 

## Test Plan

I update the CustomId component when name is project, it will use a new InputProjectId rather than InputId.

I updated the unit tests, and I verified that they all passed successfully.

before
before

after
after

## Related PRs and Issues

- https://github.com/appwrite/appwrite/issues/6279

(Write your answer here.)

### DIFF
--- a/src/lib/elements/forms/inputProjectId.svelte
+++ b/src/lib/elements/forms/inputProjectId.svelte
@@ -1,0 +1,60 @@
+<script lang="ts">
+    import { onMount } from 'svelte';
+    import { FormItem } from '.';
+    import TextCounter from './textCounter.svelte';
+
+    export let value = '';
+    export let autofocus = true;
+
+    let element: HTMLInputElement;
+    let icon = 'info';
+    const pattern = String.raw`^[a-z0-9][a-z0-9-]{1,35}$`;
+
+    onMount(() => {
+        if (element && autofocus) {
+            element.focus();
+        }
+    });
+
+    const handleInvalid = (event: Event) => {
+        event.preventDefault();
+
+        if (element.validity.patternMismatch) {
+            icon = 'exclamation';
+            return;
+        }
+    };
+
+    $: if (value) {
+        icon = 'info';
+    }
+</script>
+
+<FormItem>
+    <div class="input-text-wrapper">
+        <input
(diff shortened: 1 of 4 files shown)


src/lib/elements/forms/inputProjectId.svelte
```diff
@@ -0,0 +1,60 @@
...
+<div
+    class="u-flex u-gap-4 u-margin-block-start-8 u-small"
+    class:u-color-text-warning={icon === 'exclamation'}>
+    <span
+        class:icon-info={icon === 'info'}
+        class:icon-exclamation={icon === 'exclamation'}
+        class="u-cross-center u-line-height-1 u-color-text-gray"
+        aria-hidden="true" />
+    <span class="text u-line-height-1-5">
+        Allowed characters: lowercase alphanumeric character and non-leading hyphen
```
stnguyen90, line 58: it starts with "Allowed characters" so no need to say character again.

```suggestion
        Allowed characters: lowercase alphanumeric and non-leading hyphen
```

src/lib/elements/forms/inputProjectId.svelte
```diff
@@ -0,0 +1,60 @@
...
+    import { onMount } from 'svelte';
+    import { FormItem } from '.';
+    import TextCounter from './textCounter.svelte';
+
+    export let value = '';
+    export let autofocus = true;
+
+    let element: HTMLInputElement;
+    let icon = 'info';
+    const pattern = String.raw`^[a-z0-9][a-z0-9-]{1,35}$`;
```
stnguyen90, line 11: Something seems wrong with this pattern when it actually gets built. in the browser console logs i see:

> Pattern attribute value ^[a-z0-9][a-z0-9-]{1,35}$ is not a valid regular expression: Uncaught SyntaxError: Invalid regular expression: /^[a-z0-9][a-z0-9-]{1,35}$/v: Invalid character class

And it doesn't look like the client-side validation works because the network request is sent despite the invalid id:

image

stnguyen90 (changes requested):
Great PR! 🤯 We left some comments during the review, please check them out.

PR #609, opened 2023-10-30
Refactor filters to be generic

stnguyen90:
## What does this PR do?

- Refactor filters so that they can be used on pages other than the documents page.
- Fix [filtering on falsy values](https://github.com/appwrite/appwrite/issues/6979)
- Fix filteriing on booleans

## Test Plan

Manual

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/6979

Yes

### DIFF
--- a/src/lib/components/filters/content.svelte
+++ b/src/lib/components/filters/content.svelte
@@ -135,6 +135,16 @@
             <div class="u-margin-block-start-8">
                 {#if column.type === 'integer' || column.type === 'double'}
                     <InputNumber id="value" bind:value placeholder="Enter value" />
+                {:else if column.type === 'boolean'}
+                    <InputSelect
+                        id="value"
+                        placeholder="Select a value"
+                        required={true}
+                        options={[
+                            { label: 'True', value: true },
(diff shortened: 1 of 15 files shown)


src/lib/components/filters/content.svelte
```diff
@@ -98,7 +102,10 @@
     // This Map is keyed by tags, and has a query as the value
     function addFilter() {
         if (column && operator) {
```
TGlide, line 104: To avoid nested ifs, maybe we can do an early return here? E.g.

```suggestion
        if (!column || !operator) return
```

src/lib/components/filters/content.svelte
```diff
@@ -108,7 +115,7 @@
     }
 
     $: isDisabled = (function getDisabled() {
-        return !operator || (!operator?.hideInput && !value);
+        return !operator;
     })();
```
TGlide, line 119: We can refactor to `$: isDisabled = !operator` I'd say

TGlide (approved):

TorstenDittmann (approved):

PR #621, opened 2023-11-15
fix: deployments empty state create

TorstenDittmann:
## What does this PR do?

Bildschirmfoto 2023-11-15 um 17 19 53
Bildschirmfoto 2023-11-15 um 17 19 12

## Test Plan

- manual

✅

### DIFF
--- a/src/lib/components/empty.svelte
+++ b/src/lib/components/empty.svelte
@@ -7,6 +7,7 @@
     import { trackEvent } from '$lib/actions/analytics';
 
     export let single = false;
+    export let noMedia = false;
     export let target: string = null;
     export let href: string = null;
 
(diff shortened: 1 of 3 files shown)


src/lib/components/empty.svelte
```diff
@@ -23,17 +24,19 @@
...
-                on:click|preventDefault
-                on:click={track}
-                aria-label="create {target}">
-                {#if $app.themeInUse === 'dark'}
-                    <img src={EmptyDark} alt="create" aria-hidden="true" width="376" />
-                {:else}
-                    <img src={EmptyLight} alt="create" aria-hidden="true" width="376" />
-                {/if}
-            </button>
+            {#if !noMedia}
```
TGlide, line 27: This can be confusing, a potential improvement could be having a `hasMedia` variable instead

TGlide (approved):

src/lib/components/empty.svelte
TorstenDittmann: > This can be confusing, a potential improvement could be having a `hasMedia` variable instead

this is the only case without showing a media - so the exception. Otherwise it would need to have `hasMedia` in 10+ places 👍🏻

TGlide: I see!

PR #625, opened 2023-11-20
Fix the allowed extensions for uploading .env files

stnguyen90:
## What does this PR do?

It should not include the leading dot because the InputFile component adds the leading prefix so including the leading dot results in an extra dot like:

image

## Test Plan

Before the change, .env files were not selectable:

Screen Shot 2023-11-20 at 10 06 07 AM

and trying to drag a .env file into the modal results in:

image

After the change, I can now select any file including .env files.

image

## Related PRs and Issues

None

Yes

### DIFF
--- a/src/routes/console/project-[project]/uploadVariablesModal.svelte
+++ b/src/routes/console/project-[project]/uploadVariablesModal.svelte
@@ -85,7 +85,7 @@
         {/if}
     </div>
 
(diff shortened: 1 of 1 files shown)


TorstenDittmann:
I think we should just allow all extensions. There is also cases of .env.production and co.

eldadfux (approved):

PR #677, opened 2023-12-21
fix: upgrade flow

ArmanNik:


eldadfux (approved):

src/routes/console/wizard/cloudOrganizationChangeTier/inviteMembers.svelte
```diff
@@ -12,12 +12,24 @@
...
 
     const plan = $plansInfo.plans.find((p) => p.$id === $changeOrganizationTier.billingPlan);
 
     let email: string;
 
+    onMount(async () => {
+        const members = await sdk.forConsole.teams.listMemberships($changeOrganizationTier.id);
+        if (members.total) {
+            $changeOrganizationTier.collaborators = members.memberships.map((m) => {
+                if (m.userEmail !== $user.email) return m.userEmail;
```
TGlide, line 28: If they are the same, it will return undefined. Shouldn't you filter those out?

TGlide (approved):

PR #681, opened 2023-12-21
fix: limits for membership

ArmanNik:


src/lib/stores/billing.ts
```diff
@@ -183,18 +183,23 @@ export function checkForTrialEnding(org: Organization) {
     }
 }
 
-export function checkForUsageLimit(org: Organization) {
+export async function checkForUsageLimit(org: Organization) {
     if (!org?.billingLimits) {
         readOnly.set(false);
         return;
     }
     const { bandwidth, documents, executions, storage, users } = org.billingLimits;
```
christyjacob4, line 191:
```suggestion
    const { bandwidth, documents, executions, storage, users } = org?.billingLimits;
```
Don't we need conditional chaining in all the places ?

src/routes/console/wizard/cloudOrganizationChangeTier/inviteMembers.svelte
```diff
@@ -16,13 +16,14 @@
...
     import { user } from '$lib/stores/user';
+    import { organization } from '$lib/stores/organization';
 
     const plan = $plansInfo.plans.find((p) => p.$id === $changeOrganizationTier.billingPlan);
 
     let email: string;
 
     onMount(async () => {
-        const members = await sdk.forConsole.teams.listMemberships($changeOrganizationTier.id);
+        const members = await sdk.forConsole.teams.listMemberships($organization.$id);
```
christyjacob4, line 26: conditional chaining ?

src/lib/stores/billing.ts
```diff
@@ -183,18 +183,23 @@ export function checkForTrialEnding(org: Organization) {
...
+    const members = await sdk.forConsole.teams.listMemberships(org.$id);
+    const plan = get(plansInfo).plans.find((plan) => plan.$id === org.billingPlan);
```
stnguyen90, line 193: Can we store plansInfo as a Map instead of an Array so that it's easier to lookup stuff? If so, we can definitely do this later rather than now.

src/routes/console/wizard/cloudOrganizationChangeTier/choosePlan.svelte
```diff
@@ -57,11 +57,12 @@
...
-            $changeOrganizationTier.limitOverflow.storage > 0 ||
-            $changeOrganizationTier.limitOverflow.users > 0 ||
-            $changeOrganizationTier.limitOverflow.executions > 0 ||
-            $changeOrganizationTier.limitOverflow.members > 0
+            ($changeOrganizationTier.limitOverflow.bandwidth > 0 ||
+                $changeOrganizationTier.limitOverflow.storage > 0 ||
+                $changeOrganizationTier.limitOverflow.users > 0 ||
+                $changeOrganizationTier.limitOverflow.executions > 0 ||
+                $changeOrganizationTier.limitOverflow.members > 0) &&
+            $changeOrganizationTier.billingPlan === 'tier-0'
```
stnguyen90, line 65: Can we use an enum/constants for billing plans instead of magic strings? If so, we can definitely do this later rather than now.

src/routes/console/wizard/cloudOrganizationChangeTier/planExcess.svelte
```diff
@@ -59,12 +59,14 @@
...
             <Button
                 text
+                external
                 href="https://appwrite.io/docs/advanced/platform/starter#reaching-resource-limits">
                 Learn more
             </Button>
         {:else if currentTier === 'tier-1'}
             <Button
                 text
+                external
```
stnguyen90, line 69: Yay for this! I need it in Messaging 😅

src/routes/console/wizard/cloudOrganizationChangeTier/inviteMembers.svelte
ArmanNik: Not needed, if we don't have the org you can't get to this place

src/routes/console/wizard/cloudOrganizationChangeTier/choosePlan.svelte
ArmanNik: Makes sense as a future refactor 👍

src/lib/stores/billing.ts
ArmanNik: Makes sense as a future refactor 👍

src/lib/stores/billing.ts
```diff
@@ -183,18 +183,23 @@ export function checkForTrialEnding(org: Organization) {
...
 }
 
-export function checkForUsageLimit(org: Organization) {
+export async function checkForUsageLimit(org: Organization) {
     if (!org?.billingLimits) {
         readOnly.set(false);
         return;
     }
-    const { bandwidth, documents, executions, storage, users } = org.billingLimits;
+    const { bandwidth, documents, executions, storage, users } = org?.billingLimits;
```
TGlide, line 191:
```suggestion
    const { bandwidth, documents, executions, storage, users } = org?.billingLimits ?? {}
```

src/lib/stores/billing.ts
```diff
@@ -183,18 +183,23 @@ export function checkForTrialEnding(org: Organization) {
...
+    const members = await sdk.forConsole.teams.listMemberships(org.$id);
+    const plan = get(plansInfo).plans.find((plan) => plan.$id === org.billingPlan);
+    const membersOverflow =
+        members?.total > plan.members ? members.total - (plan.members || Infinity) : 0;
```
TGlide, line 195: if plan.members = 0, this will make it minus infinity. Isn't it better to do:
`members?.total > plan.members ? members.total - (plan.members || members.total) : 0;`

TGlide (approved):

PR #695, opened 2023-12-29
fix:  limit not being updated on members deletion

ArmanNik:
Also fixed page invalidation

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -8,6 +8,7 @@ export enum Dependencies {
     ADDRESS = 'dependency:address',
     PAYMENT_METHODS = 'dependency:paymentMethods',
     ORGANIZATION = 'dependency:organization',
+    MEMBERS = 'dependency:members',
     PROJECT = 'dependency:project',
     PROJECT_VARIABLES = 'dependency:project_variables',
     PROJECT_INSTALLATIONS = 'dependency:project_installations',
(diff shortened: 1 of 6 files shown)


src/lib/constants.ts
```diff
@@ -8,6 +8,7 @@ export enum Dependencies {
     ADDRESS = 'dependency:address',
     PAYMENT_METHODS = 'dependency:paymentMethods',
     ORGANIZATION = 'dependency:organization',
+    MEMBERS = 'dependency:members',
```
stnguyen90, line 11: Do we need to make sure to invalidate this when adding a member too?

src/routes/console/organization-[organization]/deleteMember.svelte
```diff
@@ -25,6 +29,11 @@
             } else {
                 dispatch('deleted');
             }
+            invalidate(Dependencies.ACCOUNT);
```
Comment (stnguyen90) [inline src/routes/console/organization-[organization]/deleteMember.svelte:32]: Why do we need to invalidate account?

Comment (ArmanNik) [reply src/routes/console/organization-[organization]/deleteMember.svelte:32]: Not sure but it was there and I'm afraid to remove it :rofl:

src/lib/constants.ts
ArmanNik: Fixed :)

stnguyen90 (approved):

TorstenDittmann (approved):

PR #697, opened 2023-12-29
refactor: billing endpoints

ArmanNik:


src/lib/stores/billing.ts
```diff
@@ -211,7 +211,10 @@ export async function checkPaymentAuthorizationRequired(org: Organization) {
 
 export async function paymentExpired(org: Organization) {
     if (!org?.paymentMethodId) return;
-    const payment = await sdk.forConsole.billing.getPaymentMethod(org.paymentMethodId);
+    const payment = await sdk.forConsole.billing.getOrganizationPaymentMethod(
```
lohanidamodar, line 214: Where is this method used?

lohanidamodar (changes requested):
Left a query. Also where are the similar changes for billingAddresses?

src/lib/stores/billing.ts
ArmanNik: We use this on load to check if there are any expired payments

lohanidamodar (changes requested):
We need similar changes applied for `billingAddress` as well. I don't see those.

lohanidamodar (changes requested):
Hey while loading billing address in the organizations billing section, we should always use the endpoint from organization instead from account
image

I don't yet see that change?

ArmanNik:
@lohanidamodar there's no change because we only fetch all the addresses never only one. 

If there's also a new endpoint for listing addresses, I wasn't informed and it's not mentioned in the linear task either :thinking:

lohanidamodar:
> @lohanidamodar there's no change because we only fetch all the addresses never only one.
> 
> If there's also a new endpoint for listing addresses, I wasn't informed and it's not mentioned in the linear task either 🤔

How do you display the Organization's assigned billing address? don't you need to fetch the billing address by ID when people are looking at the billing tab for the organization?

lohanidamodar:
> > > @lohanidamodar there's no change because we only fetch all the addresses never only one.
> > > If there's also a new endpoint for listing addresses, I wasn't informed and it's not mentioned in the linear task either 🤔
> > 
> > 
> > How do you display the Organization's assigned billing address? don't you need to fetch the billing address by ID when people are looking at the billing tab for the organization?
> 
> We are already fetching the addressList when we load the page so I'm just finding it among them
> 
> ```
> $addressList?.billingAddresses?.find(
>         (address) => address.$id === $organization?.billingAddressId
>     );
> ```

As we discussed, lets use the new endpoint through `/organizations/billing-addresses/:billingAddressId` to fetch the organization's billing address.

lohanidamodar (commented):
Looks good. However I think this should target `cloud` branch.

lohanidamodar (approved):

PR #708, opened 2024-01-04
Fix button spacing

baskrahmer:
Fixes that there is currently no spacing between the "Build logs" and "Redeploy" buttons for functions.


stnguyen90:
Before:

image

After:

image

baskrahmer:
@stnguyen90 Ah my bad! I see now that it renders differently for me because I use Darkeader :facepalm: thanks for checking

stnguyen90:
@baskrahmer, oh I was just attaching screenshots to help reviewers see what the difference is. That said, Carla said:

> I think the spacing might look off because of the padding around the two text buttons next to each other. 

She's going to look into this to see if we should make any change to the spacing.

PR #712, opened 2024-01-09
Add a route to show user targets

stnguyen90:
## What does this PR do?

Add a route to show user targets

There's some code for creating a target, but it must go through design first.

## Test Plan

Manual

## Related PRs and Issues

Parent: 

* https://github.com/appwrite/console/pull/704

Child:

* https://github.com/appwrite/console/pull/716
* https://github.com/appwrite/console/pull/734

Yes


stnguyen90:
@chenparnasa, you okay with going forward with this, for now?

src/routes/console/project-[project]/auth/user-[user]/targets/+page.ts
```diff
@@ -0,0 +1,83 @@
...
+                    `${sdk.forProject.client.config.endpoint}/messaging/providers/${target.providerId}`
+                ),
+                {
+                    'X-Appwrite-Project': sdk.forProject.client.config.project,
+                    'content-type': 'application/json',
+                    'X-Appwrite-Mode': 'admin'
+                }
+            );
+        }
+    });
```
Comment (stnguyen90) [inline src/routes/console/project-[project]/auth/user-[user]/targets/+page.ts:64]: Fetching providers here because Jake said:

> Any way the API does this will be exceedingly heavy-handed, since there are likely to be only a few providers but many targets, can we use a memory-map and just fetch ones into it as they're missing while iterating targets?

PR #724, opened 2024-01-15
fix: nps feedback design review

ArmanNik:


TorstenDittmann:
- hover buttons needs fix
- mobile version has overflow

TorstenDittmann (approved):

PR #728, opened 2024-01-16
Add forceRedirect handling

TGlide:
## What does this PR do?

Adds redirect handling when passing in `forceRedirect`. Needed for init


benmccann:
Why remove the pnpm lockfile?

ArmanNik (approved):
We might want to use a clearer name that includes the word `external`, but the code looks good :+1:

TGlide:
> Why remove the pnpm lockfile?

Unfortunately we use npm still. We had both npm and pnpm's lockfile as a mistake.

PR #725, opened 2024-01-15
chore: upgrade prettier

benmccann:
## What does this PR do?

Upgrade prettier

## Test Plan

I ran `pnpm format` to check that it worked okay. I added support for the upgrade in the SvelteKit repo and have made this same change in tons of other repos, so I'm pretty confident in the pattern

## Related PRs and Issues

N/A

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

No

PR #733, opened 2024-01-17
upgrade to SvelteKit 2

benmccann:
## What does this PR do?

Upgrade to SvelteKit 2

## Test Plan

I actually couldn't get this project to build on `main`. This PR fixes it. This is my first PR to the project though, so I wouldn't mind an extra set of eyes

## Related PRs and Issues

I'm also upgrading `prettier` in https://github.com/appwrite/console/pull/725

no


TorstenDittmann:
hey @benmccann 👋🏻 

I'll be able to take some time to review it tomorrow 😄

ArmanNik (changes requested):
Hey @benmccann thank you for this PR :tada: 
Looks good! I think there are still some `goto` that lead to external sources like 
https://github.com/appwrite/console/blob/e300174f6b810c236d78ab4b57929dfbd1ec88ef/src/routes/auth/magic-url/%2Bpage.svelte#L21
And a couple other places where we redirect to `appwrite-callback-` :+1:

benmccann:
Thanks for the reviews! I've updated those `goto` calls. I didn't invoke `resolve` in those calls as was being done previously since it's now invoking the browser API directly and there's not really a way for it to succeed or fail that we would worry about and you're no longer in the application once `window.location` has been updated

ArmanNik (approved):

Issue #722, opened 2024-01-13
🚀 Feature: improve login page core web vitals


benmccann:
### 🔖 Feature description

Prerender the login page for improved core web vitals

### 🎤 Pitch

While investigating SvelteKit applications in the wild, I found that AppWrite consoles are written in SvelteKit and fail core web vitals. E.g. here's one such URL: https://auth.lucas-cm.com.br/login

A big contributor to this is that [SSR is disabled](https://github.com/appwrite/console/blob/92e7f4780738c16b0e403b69635b9d1d14041ce2/src/routes/%2Blayout.ts#L10), so just a blank page loads initially.

This could best be avoided by prerendering the login page.

(Alternately, the largest image on the page could be preloaded and you could render some image, background image, or text while the JavaScript is being fetched rather than leaving the page blank, but this would be more work, a worse user experience, and probably still worse on core web vitals)

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

benmccann:
It appears to me that this is deployed using Docker, so the best solution is probably to switch to `adapter-node` and remove `ssr = false`.

@TorstenDittmann I see `ssr = false` was added in https://github.com/appwrite/console/commit/c8065121fdb72095a44daf66c12c0c89fc208b4f, but I'm not sure why. Do you see any issues with removing `ssr = false` and switching to `adapter-node`?

eldadfux:
Starting today this repo is archived. Please raise any issues with the new Appwrite console on the main platform repo: https://github.com/appwrite/appwrite

PR #738, opened 2024-01-19
prerender login page for better performance

benmccann:
## What does this PR do?

Prerenders the login page for much better performance in loading it

## Test Plan

I am unable to test this as the build is broken for me. That would be fixed by https://github.com/appwrite/console/pull/733

## Related PRs and Issues

Closes https://github.com/appwrite/console/issues/722

no

### DIFF
--- a/src/routes/login/+page.js
+++ b/src/routes/login/+page.js
@@ -1,0 +1,2 @@
+export const ssr = true;
+export const prerender = true;


benmccann:
It would be possible to make this work, but I just saw that this is deployed using Docker, which means that re-enabling SSR is probably a better solution

PR #740, opened 2024-01-23
Partial prerendering of public pages for better core web vitals

benmccann:
Closes https://github.com/appwrite/console/issues/722

Previously it would display a blank white screen, followed by a loading widget, followed by the actual content.

Now it displays the left half of the content with the logo and title immediately and then the right half of the content with page-specific contents is filled in later.

Note that turning on SSR would be a more performant solution for the rest of the pages and not just the public facing ones I'm tweaking here. However, I understand there's a desire to run only a single process (PHP server) and avoid the operational overhead of introducing a Node server, so I've sent this PR as a solution instead.

### DIFF
--- a/src/lib/stores/app.ts
+++ b/src/lib/stores/app.ts
@@ -6,10 +6,12 @@ export type AppStore = {
     themeInUse: 'light' | 'dark';
 };
 
-export const app = writable<AppStore>({
+const defaultValue = {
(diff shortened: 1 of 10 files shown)


TorstenDittmann:
We have moved to a Node.js container now and we are taking care of it for next release 💪🏻

PR #771, opened 2024-02-06
Use enum and SDK methods

stnguyen90:
## What does this PR do?

* Use enums from the console SDK whenever possible
* Remove unnecessary backup
* Fix listInvoices() queries type
* Add messaging service to sdkForProject
* Replace client.call() with the respective SDK methods

Some svelte diagnostic checks still don't pass because of problems with the SDK.

## Test Plan

Manual

## Related PRs and Issues

None

Yes


stnguyen90:
@TorstenDittmann, sorry for the huge PR. The [enum change](https://github.com/appwrite/console/pull/771/commits/f9a4f9850a14eb8854ea0071636b74b591fa6dff) was really big.

src/lib/components/creditCardBrandImage.svelte
```diff
@@ -1,12 +1,14 @@
...
     import { sdk } from '$lib/stores/sdk';
+    import { CreditCard } from '@appwrite.io/console';
 
     export let brand: string;
     export let width = 23;
     export let height = 16;
 
     function getCreditCardImage(brand: string, width = 46, height = 32) {
-        if (!brand) return '';
+        if (!isValueOfStringEnum(CreditCard, brand)) return '';
```
abnegate, line 11: Is it possible to change the type of the method parameter instead so the method can't be called with an invalid value?

stnguyen90: I definitely think that's the ideal approach. I was hesitant because there are a lot of places where the `CreditCardBrandImage` component is used and the change would be a little big. How about we tackle it in a separate PR?

PR #862, opened 2024-02-23
Update 'envVar' to 'environment variable' for migrations and remove Firebase OAuth stuff

PineappleIOnic:
## What does this PR do?

Migrations lib has switched `envVar` to `environment variable` which also trickles down to the API Controller, this PR updates that.

Yes


pnpm-lock.yaml
TorstenDittmann, line 1: leftover

src/lib/stores/migration.ts
```diff
@@ -42,7 +42,7 @@ const resources = [
     'file',
     'bucket',
     'function',
-    'envVar',
+    'environment variable',
```
TorstenDittmann, line 45: just out of curioisity, why aren't we sticking to space instead of dash or so?

TorstenDittmann (approved):

PR #878, opened 2024-02-25
Fix: 2FA autofocus

Meldiron:
## What does this PR do?

Autofocus did not work in 2FA login screen. Now it does

## Test Plan

- [x] Manual QA

## Related PRs and Issues

x

yes

### DIFF
--- a/src/lib/elements/forms/inputDigits.svelte
+++ b/src/lib/elements/forms/inputDigits.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
-    import { onMount } from 'svelte';
+    import { onMount, tick } from 'svelte';
     import { FormItem } from '.';
     import { createPinInput, melt } from '@melt-ui/svelte';
+    import Body from '../table/body.svelte';
 
     export let length: number = 6;
     export let value: string = '';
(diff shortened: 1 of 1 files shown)


src/lib/elements/forms/inputDigits.svelte
```diff
@@ -24,10 +25,30 @@
...
+
+        await tick();
+
+        setTimeout(() => {
+            focusInput(i + 1);
+        }, 50);
+    }
+
+    onMount(() => {
+        focusInput();
```
TorstenDittmann, line 51: can you check if this would work?

```js
let interval = setInterval(() => {
    const element = element.querySelector('input');
    if (element) {
        element.focus();
        clearInterval(interval);
    }
}, 50);

Meldiron: Doesnt work. `if (element) {` is true on first iteration. Trust me, no elegant solution here. All values in HTML seem to say what it should, but then Svelte Kit must interfere to unfocus. And it starts to work if `ssr = true` in `layout.page`.
I even tried `document.hasFocus` and that was true as expected, but then few ticks later it became false.

PR #914, opened 2024-03-02
Add tab to show user's identities

stnguyen90:
## What does this PR do?

Show users' identities and allow deleting them.

## Test Plan

Manual

image

## Related PRs and Issues

Child:

* https://github.com/appwrite/console/pull/921

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes

PR #921, opened 2024-03-05
Add account identities

stnguyen90:
## What does this PR do?

Add a card so that developers can see their identities.

## Test Plan

Manual

## Related PRs and Issues

Parent:

* #914 

Yes


src/routes/console/account/store.ts
```diff
@@ -6,3 +6,7 @@ export const factors = derived(
     page,
     ($page) => $page.data.factors as Models.MfaFactors & { recoveryCode: boolean }
 );
+export const identities = derived(
+    page,
+    ($page) => $page.data.identities.identities as Models.Identity[]
```
TorstenDittmann, line 11: seeing this, just as a thought.

we have some places with these weird notations `documents.documents`, `users.users`, etc.

we could think about improving some of those :-) not important for this PR

TorstenDittmann (approved):

src/routes/console/account/store.ts
stnguyen90: Ya good point. We should change them to something like `$page.data.identityList.identities` to match the type `IdentityList`, ya?

TorstenDittmann: yeah, something like that 👍🏻

PR #994, opened 2024-03-29
chore: remove log rocket

ArmanNik:


ArmanNik:
- Not sure it's worth the wor. It's just an extra `data-private` on some HTML tags and we might use it again if we implement another solution like sentry or posthog
- Until cloud is on 1.5.x we are going to need two branches, we sync them regularly though
- Should be fixed not :+1:

eldadfux (approved):

PR #1002, opened 2024-04-02
fix: upload wizard not closing

ArmanNik:
When uploading a larger file like attached here, it feel like the window is frozen, at some point the submit button becomes enabled and I can submit the file again and again.

It's okay to not await the `createFile()` call because there are other visual indicators for the upload like the upload progress and the success notification.


src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte
```diff
@@ -29,7 +29,7 @@
 
     async function create() {
         try {
-            await uploader.uploadFile(
+            uploader.uploadFile(
```
Comment (stnguyen90) [inline src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte:32]: The problem with doing this is:

1. the list of files doesn't update after the file is uploaded
2. the success notification is inaccurate

Maybe it would be better to move:

```
            createFile.reset();
            wizard.hide();
```

up before the `await uploadFile()`?

Comment (ArmanNik) [reply src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte:32]: We can't do the reset before the upload otherwise there's nothing to upload :sweat_smile: 
This means that the order needs to be the same, and we can't await the upload otherwise if the dev tries to upload something else, there's still going to be the previous file selected

Comment (ArmanNik) [reply src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte:32]: I'll ping Carla about the notification message

TorstenDittmann (approved):

PR #1003, opened 2024-04-02
fix: attributes creation racing condition

ArmanNik:
Fixes: https://github.com/appwrite/appwrite/issues/7800


stnguyen90:
@ArmanNik, this issue should be fixed now. Have you tested?

PR #1006, opened 2024-04-03
feat: add downgrade alerts

ArmanNik:
Add an inline alert to the change plan/downgrade confirmation step in the modal to make it more clear that plan subscriptions will end after payment of their last invoice/completion of the current billing cycle.


heyCarla (approved):
🎉

stnguyen90 (approved):

Issue #1030, opened 2024-04-12
🐛 Bug Report: Cannot Delete Org.


ItzNotABug:
### 👟 Reproduction steps

1. Go to Console > Create a Test Org.
2. Go to the Settings Tab of that Org.
3. Click on Delete > nothing happens

See -

https://github.com/appwrite/console/assets/20625965/9f8cb5c7-cfed-48fa-90e5-592e38ff65d4

---
I'd like to be assigned to work on this.

### 👍 Expected behavior

Show Delete Modal.

### 👎 Actual Behavior

Nothing Happens.

### 🎲 Appwrite version

Version 1.5.x

### 💻 Operating system

MacOS

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

PR #1032, opened 2024-04-14
Fix: Show Delete Org. Modal on Self Hosted Instances

ItzNotABug:
## What does this PR do?

Show organization deletion modal on self hosted instances.

## Test Plan

Manual.

https://github.com/appwrite/console/assets/20625965/a2e38714-2085-47b5-a2ec-0fc70e3ecb30

## Related PRs and Issues

* #1030

Yes.

### DIFF
--- a/src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
+++ b/src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
@@ -51,7 +51,7 @@
         }
     }
 
-    $: upcomingInvoice = invoices.invoices.find((i) => i.status === 'upcoming' && i.amount > 0);
+    $: upcomingInvoice = invoices?.invoices.find((i) => i.status === 'upcoming' && i.amount > 0);
 </script>
 
 <Modal
--- a/src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
+++ b/src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
@@ -37,6 +37,8 @@
                         : $organizationList.teams[0].$id;
 
                 await goto(`${base}/console/organization-${teamId}`);
+            } else if ($organizationList?.total == 1) {
+                await goto(`${base}/console/organization-${$organizationList.teams[0].$id}`);
             } else {
                 await goto(`${base}/console/onboarding`);
             }
--- a/src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
+++ b/src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
@@ -73,16 +73,9 @@
         </p>
     {:else}
         <p data-private>
-            Are you sure you want to delete <b>{$organization.name}</b>?
-            {#if $projects.total > 0}
-                All projects ({$projects.total}) and data associated with this organization will be
-                deleted.
-            {:else}
-                There are <b>no projects</b> associated with this organization.
-            {/if}
+            Are you sure you want to delete <b>{$organization.name}</b>? All projects ({$projects.total})
+            and data associated with this organization will be deleted. This action is irreversible.
         </p>
-
-        <p><b>This action is irreversible.</b></p>
     {/if}
     <svelte:fragment slot="footer">
         <Button text on:click={() => (showDelete = false)}>Cancel</Button>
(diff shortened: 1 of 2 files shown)


src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
```diff
@@ -37,6 +37,8 @@
                         : $organizationList.teams[0].$id;
 
                 await goto(`${base}/console/organization-${teamId}`);
+            } else if ($organizationList?.total == 1) {
+                await goto(`${base}/console/organization-${$organizationList.teams[0].$id}`);
```
Comment (Meldiron) [inline src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte:41]: I would avoid this change, I don't think the design team would approve. 
A developer might be deleting a project with the intention of creating a new one, simply trying to wipe it and re-create it from appwrite.json, for instance.

Redirecting him to a project would stand in the way.

In other scenarios, it could also cause confusion, making the developer think it's same project, just dashboard. He would think something went wrong, will refresh page and try to delete again - causing him to delete a whole different project.

src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
```diff
@@ -51,7 +53,7 @@
         }
     }
 
-    $: upcomingInvoice = invoices.invoices.find((i) => i.status === 'upcoming' && i.amount > 0);
+    $: upcomingInvoice = invoices?.invoices.find((i) => i.status === 'upcoming' && i.amount > 0);
```
Comment (Meldiron) [inline src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte:56]:
```suggestion
    $: upcomingInvoice = invoices?.invoices?.find((i) => i.status === 'upcoming' && i.amount > 0);
```

I think a second `?` will be necessary, otherwise, it becomes `null.find()`.  I think, I might be totally wrong.

src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
```diff
@@ -73,9 +75,16 @@
...
+            Are you sure you want to delete <b>{$organization.name}</b>?
+            {#if $projects.total > 0}
+                All projects ({$projects.total}) and data associated with this organization will be
+                deleted.
+            {:else}
+                There are <b>no projects</b> associated with this organization.
+            {/if}
         </p>
+
+        <p><b>This action is irreversible.</b></p>
```
Comment (Meldiron) [inline src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte:87]: cc design team about this change.

Currently, modal just has a copy, and says 0 projects in case it's 0.

Suggestion here is to have separate copy for when there is 0 projects. Inteiton, I believe, is to scare the developer a bit less in such scenario.

I personally like this change, but I would love to hear from design team to see if we can make it **more** scare for when there are projects, compared to when there are 0 proejcts.

src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
Comment (ItzNotABug) [reply src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte:41]: I'd see an onboarding page every time on an org. deletion, thought it wasn't expected if there's already atleast one organisation available. But the reasoning regarding the confusion makes sense. I'll roll back this change.

src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
Comment (ItzNotABug) [reply src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte:56]: I don't think so, `Cloud` always returns an empty list if there are no invoices and returns `undefined` if on self-hosted.

---
Here's an actual return from the api -
URL - `https://cloud.appwrite.io/v1/organizations/{:orgId}/invoices?organizationId={:orgId}`
Response -
```json
{
  "total": 0,
  "invoices": []
}
```

Let me know if I should anyway add a second `?`.

src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
Comment (ItzNotABug) [reply src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte:87]: Thanks for the elaboration. I'll wait in the meantime for design team's feedback.

Note: In order to make this a bit more scary, we could increate the text size of `This action is irreversible`. I am not sure if using a brighter color would look good though. Current color scheme seems fine enough.

src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
Comment (ItzNotABug) [reply src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte]: Done.

src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte
Comment (Meldiron) [reply src/routes/console/organization-[organization]/settings/deleteOrganizationModal.svelte:54]: You are right, seems to be fine as it is.
CleanShot 2024-04-16 at 10 57 22@2x

PR #1045, opened 2024-04-18
Track source

lohanidamodar:
## What does this PR do?

- Track referral source

## Test Plan

## Related PRs and Issues

YES

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -21,7 +21,50 @@
         window.VERCEL_ANALYTICS_ID = import.meta.env.VERCEL_ANALYTICS_ID?.toString() ?? false;
     }
 
+    let getCookie = (name) => {
+        let nameEQ = name + '=';
+        let ca = document.cookie.split(';');
+        console.log(ca);
+        for (var i = 0; i < ca.length; i++) {
+            let c = ca[i];
+            while (c.charAt(0) == ' ') c = c.substring(1, c.length);
+            if (c.indexOf(nameEQ) == 0) return c.substring(nameEQ.length, c.length);
+        }
+        return null;
+    };
+
+    let setCookie = (name, value) => {
+        document.cookie =
+            name +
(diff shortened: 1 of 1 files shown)


src/routes/+layout.svelte
```diff
@@ -16,12 +16,42 @@
...
     import Consent from '$lib/components/consent.svelte';
+    import { sdk } from '$lib/stores/sdk';
 
     if (browser) {
         window.VERCEL_ANALYTICS_ID = import.meta.env.VERCEL_ANALYTICS_ID?.toString() ?? false;
     }
 
     onMount(async () => {
+        // handle sources
+        if (browser && isCloud) {
```
TorstenDittmann, line 27: we dont need to check for browser, onmount only runs on browser anyway 👍🏻 

```suggestion
        if (isCloud) {
```

src/routes/+layout.svelte
TorstenDittmann, line 1: I am a bit confused about this - why do we create a source on visit?

this can be done with plausible already without any implementation.

TorstenDittmann, line 1: this will run everytime - we shouldnt do this, only when it's not set right?

does the create source set a cookie or something with the fingerprint?

src/routes/+layout.svelte
```diff
@@ -16,12 +16,42 @@
...
+            let referrer = document.referrer.length ? document.referrer : null;
+
+            // Skip our own
+            if (referrer?.includes('//appwrite.io')) {
+                referrer = null;
+            }
+
+            if (ref || referrer || utmSource || utmCampaign || utmMedium) {
+                try {
+                    await sdk.forConsole.sources.create(
```
TorstenDittmann, line 42: we shouldnt await it 👍🏻 otherwise its blocking following logic

lohanidamodar: updated

PR #1057, opened 2024-04-29
Improve Bucket Limits Message

ItzNotABug:
## What does this PR do?

This PR fixes the ambiguous file size shown on bucket limits without specifying the type of the size limit.

## Test Plan

Manual -

Screenshot 2024-04-29 at 4 29 35 PM

## Related PRs and Issues

NA.

Yes.

### DIFF
--- a/src/routes/console/project-[project]/storage/bucket-[bucket]/+page.svelte
+++ b/src/routes/console/project-[project]/storage/bucket-[bucket]/+page.svelte
@@ -124,7 +124,7 @@
                 <p class="text">
                     <button class="link" type="button" on:click|preventDefault={upgradeMethod}
                         >Upgrade</button>
-                    for additional storage.
+                    for additional storage resources.
                 </p>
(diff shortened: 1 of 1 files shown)


heyCarla:
Thanks for catching this! I think we should follow the pattern used in other screens when we have multiple limits that apply in a given context (example below). We can say here:

**The Starter plan has limits**
2GB total file storage
50MB file upload size limit
Upgrade for additional storage resources.

Screenshot 2024-04-29 at 19 56 57

ItzNotABug:
> Thanks for catching this! I think we should follow the pattern used in other screens when we have multiple limits that apply in a given context (example below). We can say here:
> 
> **The Starter plan has limits** 2GB total file storage 50MB file upload size limit Upgrade for additional storage resources.
> 
> Screenshot 2024-04-29 at 19 56 57

Done.

stnguyen90 (approved):

PR #1062, opened 2024-05-01
Fix location of tooltip in collection table

elad2412:
Update size for<div> element to fit-content  in order for the tooltip will be in the correct location


stnguyen90 (approved):
Looks good! 

image

PR #1078, opened 2024-05-07
Update style x button size of Alert partial (reviewed with Chen)

elad2412:
Update style x button size of Alert partial to vertically align the alert icon on the left w/ the x icon on the right.


stnguyen90 (approved):

stnguyen90:
Related pink change: https://github.com/appwrite/pink/pull/194

PR #1090, opened 2024-05-15
chore: enable python runtime, bump sdk

loks0n:


pnpm-lock.yaml
stnguyen90, line 1: We don't use pnpm in the console. Would you please remove this?

stnguyen90 (changes requested):

pnpm-lock.yaml
loks0n: Done

stnguyen90 (approved):

PR #1125, opened 2024-06-04
fix: design downgrade flow

ArmanNik:


src/lib/components/billing/planComparisonBox.svelte
```diff
@@ -1,14 +1,15 @@
...
+    import { SecondaryTabs, SecondaryTabsItem } from '..';
 
     let selectedTab: 'tier-0' | 'tier-1' = 'tier-0';
+    export let downgrade = false;
 
     $: plan = $plansInfo.get(selectedTab);
 </script>
 
-<Box>
+<div class="card">
```
TorstenDittmann, line 12: Can we use the Card component here?

TorstenDittmann (approved):

src/lib/components/billing/planComparisonBox.svelte
ArmanNik: fixed :)

PR #1140, opened 2024-06-11
feat: campaign flow

ArmanNik:


ArmanNik:
1. I'll look into it
2. It's backend :innocent: 
3. We might but that's the existing flow so I didn't redo it from scratch. The modal allows us to validate the coupon easily, fetch the related data, and display the possible errors. We might have a denounced text field instead but that would require a lot of design changes and I'm not sure it's the best experience having the backed error displayed under the input  :thinking: 

@TorstenDittmann

TorstenDittmann (approved):

PR #1143, opened 2024-06-12
feat: enable button in new wizards and let form validation handle errors

ArmanNik:


src/lib/components/billing/selectPaymentMethod.svelte
```diff
@@ -85,22 +97,33 @@
...
-            </Button>
-        </div>
-    </Card>
+    <div>
+        <input
+            bind:this={input}
+            on:invalid={handleInvalid}
+            required
+            class="u-hide"
+            type="text"
```
TorstenDittmann, line 106: Maybe here we can use type=hidden

ArmanNik: image
We can't... I tested it and the hidden input doesn't have validity, which means we can't show the message

I think it's due to this:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden#validation

PR #1147, opened 2024-06-17
feat: new campaigns

ArmanNik:


static/images/campaigns/InactiveAccounts/dark.png
lohanidamodar, line 1: as campaign images are same for these two, why not add single image like coupon50-dart.png and coupon50-light.png and in the campaigns.ts for each campaign add image field?

lohanidamodar (commented):
Left a comment

static/images/campaigns/InactiveAccounts/dark.png
ArmanNik: Hmmmm not sure it's worth the change...the way it's currently implemented instead of hardcoding the image route, we get it dynamically from the campaign name

TorstenDittmann (approved):

PR #1167, opened 2024-06-28
feat: delete and cancel deployments

loks0n:


src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts
```diff
@@ -2,10 +2,17 @@ import { sdk } from '$lib/stores/sdk';
...
 
+    const deployment = await sdk.forProject.functions.getDeployment(
+        params.function,
+        params.deployment
+    );
+
     return {
-        deployment: await sdk.forProject.functions.getDeployment(params.function, params.deployment)
+        deployment,
+        activeDeployment: data.function.deployment === params.deployment ? true : false
```
Comment (ArmanNik) [inline src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts:16]: I don't think we need the ternary operator `data.function.deployment === params.deployment` is already either true or false

src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts
```diff
@@ -2,10 +2,17 @@ import { sdk } from '$lib/stores/sdk';
 import { Dependencies } from '$lib/constants';
 import type { PageLoad } from './$types';
 
-export const load: PageLoad = async ({ params, depends }) => {
+export const load: PageLoad = async ({ params, depends, parent }) => {
+    const data = await parent();
```
Comment (ArmanNik) [inline src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts:6]:
```suggestion
    const { deployment } = await parent();
```
It's easier to know what we want from the parent, and it will be highlighted as an error in the code editor if the parent does not return it

ArmanNik (changes requested):

src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts
Comment (loks0n) [reply src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts]: fixed

src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts
Comment (loks0n) [reply src/routes/console/project-[project]/functions/function-[function]/deployment-[deployment]/+page.ts]: done, but function is a reserved keyword, so renamed func

ArmanNik (approved):

PR #1193, opened 2024-07-14
Fix Stuck Uploader Progress on File Upload

ItzNotABug:
## What does this PR do?

Fixes the issue where if there was an exception during file upload, the progress component would keep showing Progress at 0%.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

Yes.


src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte
```diff
@@ -29,7 +29,7 @@
 
     async function create() {
         try {
-            uploader.uploadFile(
+            const uploadPromise = uploader.uploadFile(
```
Comment (TorstenDittmann) [inline src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte:32]: we can await here directly
```suggestion
            await uploader.uploadFile(
```

src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte
```diff
@@ -47,7 +47,11 @@
             trackEvent(Submit.FileCreate, {
                 customId: !!$createFile.id
             });
+
+            await uploadPromise;
         } catch (e) {
+            uploader.removeAtIndex(0);
```
Comment (TorstenDittmann) [inline src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte:53]: not sure we can should rely too much on the index too much. lets use the file id to remove it from the uploader 👍🏻

Comment (ItzNotABug) [reply src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte:53]: I don't think we'd have a concrete file ID in a `catch` if the upload fails. Would we?

src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte
Comment (ItzNotABug) [reply src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte:32]: Using `await` will keep the wizard open until the upload finishes. Not good in case of medium to large files.

src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte
Comment (ItzNotABug) [reply src/routes/console/project-[project]/storage/bucket-[bucket]/create-file/create.svelte]: Done.

TorstenDittmann (approved):

PR #1213, opened 2024-07-18
feat: nginx configurations

TorstenDittmann:
## What does this PR do?

- adds compression fallbacks, cache and security headers to nginx
- adds precompress to static build

✅


docker/nginx.conf
```diff
@@ -1,12 +1,34 @@
...
+    gzip_static on;
+
+    # Deny IE browsers from going into quirks mode
+    add_header X-UA-Compatible "IE=Edge";
+    # X-Frame-Options is to prevent from clickJacking attack
+    add_header X-Frame-Options SAMEORIGIN;
+    # This header enables the Cross-site scripting (XSS) filter
+    add_header X-XSS-Protection "1; mode=block;";
+    # disable content-type sniffing on some browsers.
+    add_header X-Content-Type-Options nosniff;
```
stnguyen90, line 23: Shouldn't I see these in the response headers? 🧐 

image

docker/nginx.conf
```diff
@@ -1,12 +1,34 @@
...
+
     location /console {
         root /usr/share/nginx/html;
         index index.html index.htm;
         try_files $uri /console/index.html;
+        expires $expires;
```
stnguyen90, line 29: I don't seem to see this in the response header

stnguyen90: maybe the Cache-Control one is causing a conflict?

stnguyen90 (commented):
I tried browsing to http://127.0.0.1:3000/auth/oauth2/success but it redirected me to /console. Is that expected?

TorstenDittmann:
> I tried browsing to http://127.0.0.1:3000/auth/oauth2/success but it redirected me to /console. Is that expected?

That redirect will be handled on the appwrite API side. Due to loadbalancer configurations, the console cannot act outside of the `/console` basepath. So in production it redirects to `http://127.0.0.1:3000/console/auth/oauth2/success`.

docker/nginx.conf
TorstenDittmann: Good catch, looks like adding the cache headers kinda broke them. fixed 👍🏻

docker/nginx.conf
TorstenDittmann: Only will be applied to non `text/html` requests 👍🏻

stnguyen90 (approved):

PR #1219, opened 2024-07-19
fix: design review 1.6

ArmanNik:


src/routes/(console)/project-[project]/functions/function-[function]/executions/+page.svelte
Comment (stnguyen90) [inline src/routes/(console)/project-[project]/functions/function-[function]/executions/+page.svelte:1]: Can we change the "Execute now" button to just be "Execute" since if you schedule an execution it won't actually execute now.

Comment (heyCarla) [reply src/routes/(console)/project-[project]/functions/function-[function]/executions/+page.svelte:1]: I think I got em. Can you please check?

TorstenDittmann (approved):

PR #1263, opened 2024-08-06
feat: upgrade function creation wizard

ArmanNik:


src/lib/components/dropList.svelte
```diff
@@ -31,9 +31,8 @@
         <div
             class="drop is-no-arrow"
             class:u-max-width-100-percent={fullWidth}
-            style={`${width ? `--drop-width-size-desktop:${width}rem; ` : ''} ${
-                position === 'static' ? 'position:static' : 'position:relative'
-            }`}>
+            style:--drop-width-size-desktop={width ? `${width}rem` : ''}
+            style:position>
```
TorstenDittmann, line 35: whats `style:position`? :-P

TorstenDittmann (approved):

src/lib/components/dropList.svelte
ArmanNik: We have a variable called `position` that is either `static` or `relative` so  this will be `position: relative | static` right? :thinking:

PR #1266, opened 2024-08-06
feat: use functions listtemplates

loks0n:
- refactors how function templates and runtimes are managed, moving from hardcoded data in `marketplace.ts` to API
- new stores (`templates.ts` and `runtimes.ts`) to handle template and runtime data
- updated UI components and pages to use the new stores

- [x] Manual QA

CleanShot 2024-08-15 at 15 41 51@2x


src/routes/(console)/project-[project]/functions/+layout.ts
```diff
@@ -8,15 +8,17 @@ import type { LayoutLoad } from './$types';
 export const load: LayoutLoad = async ({ depends }) => {
     depends(Dependencies.FUNCTION_INSTALLATIONS);
 
-    const [runtimesList, installations] = await Promise.all([
+    const [runtimesList, installations, templatesList] = await Promise.all([
         sdk.forProject.functions.listRuntimes(),
-        sdk.forProject.vcs.listInstallations([Query.limit(100)])
+        sdk.forProject.vcs.listInstallations([Query.limit(100)]),
+        sdk.forProject.functions.listTemplates()
```
Comment (Meldiron) [inline src/routes/(console)/project-[project]/functions/+layout.ts:14]: Default limit is 25 I think. We might want to do 100 here, until we introduce proper integration with server-side pagination and filtering.

(There is client side pagination currently)

ArmanNik (approved):

christyjacob4 (approved):

PR #1341, opened 2024-09-09
fix: show error when the file is over the max size

ArmanNik:


src/lib/elements/forms/inputFile.svelte
```diff
@@ -86,12 +97,20 @@
             const fileExtension = file.name.split('.').pop();
             return isFileExtensionAllowed(fileExtension);
         });
+        const isOverSize = maxSize
+            ? Array.from(target.files).some((file) => isFileOverSize(file))
+            : false;
```
TorstenDittmann, line 102:
```suggestion
        const isOverSize = maxSize && Array.from(target.files).some((file) => isFileOverSize(file));
```

TorstenDittmann: does this work better?

TorstenDittmann (approved):

PR #1342, opened 2024-09-09
fix: payment expired notification, add sessionstorage check

ArmanNik:
By using session storage here, the developer should only receive the notification during their current session. 

If they haven't updated their expired payment and opened a new session, they'll get the notification again.

### DIFF
--- a/src/lib/stores/billing.ts
+++ b/src/lib/stores/billing.ts
@@ -306,6 +306,8 @@ export async function paymentExpired(org: Organization) {
         org.paymentMethodId
     );
     if (!payment?.expiryYear) return;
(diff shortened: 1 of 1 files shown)


divanov11:
@stnguyen90 

***The problem:** Payment expiration error pops up on every page, even when closed. 

**Expected behavior:** Once a user gets a notification, they should be able to close and it and not see it until their next browser session.

**Solution:** Added a check for payment notification in session storage. Once a user has closed the pop-up `expiredPaymentNotification` will be set to true, indicating user has already been notified.

stnguyen90 (approved):

PR #1355, opened 2024-09-18
fix: page breaks on redirection

ArmanNik:
Fixed pages:
- auth/templated
- auth/security
- messaging/topics
- storage/files
- functions/deployments

This is probably due to a Svelte bug with stores


src/routes/(console)/project-[project]/auth/security/+page.svelte
```diff
@@ -11,12 +12,14 @@
...
 <Container>
-    <UpdateUsersLimit />
-    <UpdateSessionLength />
-    <UpdateSessionsLimit />
-    <UpdatePasswordHistory />
-    <UpdatePasswordDictionary />
-    <UpdatePersonalDataCheck />
-    <UpdateSessionAlerts />
-    <UpdateMockNumbers />
+    {#if $project?.$id}
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[project]/auth/security/+page.svelte:15]: cant we just rely on `$page.params.project`?

src/routes/(console)/project-[project]/auth/templates/+page.ts
```diff
@@ -1,9 +1,10 @@
 import { sdk } from '$lib/stores/sdk';
-import type { PageLoad } from './$types';
 
-export const load: PageLoad = async () => {
+export const load = async ({ parent }) => {
+    const { project } = await parent();
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[project]/auth/templates/+page.ts:4]: this should be available from the project layout right?

src/routes/(console)/project-[project]/auth/security/+page.svelte
Comment (ArmanNik) [reply src/routes/(console)/project-[project]/auth/security/+page.svelte:15]: Not in this case we need more data from the `project.authPasswordHistory` so we need to actually use the store

src/routes/(console)/project-[project]/auth/templates/+page.ts
Comment (ArmanNik) [reply src/routes/(console)/project-[project]/auth/templates/+page.ts:4]: I think so

src/routes/(console)/project-[project]/auth/security/+page.svelte
Comment (ArmanNik) [reply src/routes/(console)/project-[project]/auth/security/+page.svelte:15]: I guess so, but this is only two lines of codes and future proofs all future changes

Comment (TorstenDittmann) [reply src/routes/(console)/project-[project]/auth/security/+page.svelte:15]: still, the condition here potentially creates a "flicker" due to state not synced

Comment (ArmanNik) [reply src/routes/(console)/project-[project]/auth/security/+page.svelte]: removed :+1:

src/routes/(console)/project-[project]/auth/security/updatePasswordHistory.svelte
```diff
@@ -9,15 +9,14 @@
     import { sdk } from '$lib/stores/sdk';
     import { project } from '../../store';
 
-    const projectId = $project.$id;
-    let passwordHistory = $project.authPasswordHistory < 1 ? 5 : $project.authPasswordHistory;
-    let passwordHistoryEnabled = ($project.authPasswordHistory ?? 0) != 0;
+    let passwordHistory = $project?.authPasswordHistory < 1 ? 5 : $project?.authPasswordHistory;
+    let passwordHistoryEnabled = ($project?.authPasswordHistory ?? 0) != 0;
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[project]/auth/security/updatePasswordHistory.svelte:13]:
```suggestion
    let passwordHistoryEnabled = ($project?.authPasswordHistory ?? 0) !== 0;
```

src/routes/(console)/project-[project]/auth/templates/+page.svelte
```diff
@@ -105,7 +103,7 @@
         </div>
     </div>
 
-    {#if !$project.smtpEnabled}
+    {#if !data.project?.smtpEnabled}
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[project]/auth/templates/+page.svelte:106]:
```suggestion
    {#if !data.project.smtpEnabled}
```

using `data` we can be sure the project is always there

src/routes/(console)/project-[project]/auth/templates/+page.ts
Comment (TorstenDittmann) [reply src/routes/(console)/project-[project]/auth/templates/+page.ts:4]: this is still here 👀

TorstenDittmann (approved):

PR #1424, opened 2024-10-17
Fix billing issues

ernstmul:
## What does this PR do?
- don't show pending invoices
- fix redirects
- alter succes message

✅

### DIFF
--- a/src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
+++ b/src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
@@ -45,22 +45,13 @@
     onMount(request);
 
     async function request() {
-        const unfilteredInvoiceList = await sdk.forConsole.billing.listInvoices(
-            $page.params.organization,
-            [
-                Query.limit(limit),
-                Query.offset(offset),
-                Query.notEqual('from', $organization.billingCurrentInvoiceDate),
-                Query.orderDesc('$createdAt')
-            ]
-        );
-        const filteredInvoices = unfilteredInvoiceList.invoices.filter(
-            (invoice) => invoice.status !== 'pending'
-        );
-        invoiceList = {
-            invoices: filteredInvoices,
-            total: filteredInvoices.length
-        };
+        invoiceList = await sdk.forConsole.billing.listInvoices($page.params.organization, [
+            Query.limit(limit),
+            Query.offset(offset),
+            Query.notEqual('from', $organization.billingCurrentInvoiceDate),
+            Query.notEqual('status', 'pending'),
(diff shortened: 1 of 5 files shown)


src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
```diff
@@ -45,12 +45,22 @@
...
+            [
+                Query.limit(limit),
+                Query.offset(offset),
+                Query.notEqual('from', $organization.billingCurrentInvoiceDate),
+                Query.orderDesc('$createdAt')
+            ]
+        );
+        const filteredInvoices = unfilteredInvoiceList.invoices.filter(
+            (invoice) => invoice.status !== 'pending'
+        );
```
Comment (ArmanNik) [inline src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte:59]: We should be able to add an extra query to the fetch request instead of doing the filtering in the frontend right?

Something like `Query.notEqual('status', 'pending')`

src/routes/(console)/organization-[organization]/change-plan/+page.svelte
```diff
@@ -217,7 +217,7 @@
             await invalidate(Dependencies.ACCOUNT);
             await invalidate(Dependencies.ORGANIZATION);
 
-            await goto(`${base}/organization-${org.$id}`);
+            await goto(`${base}/organization-${org.$id}/billing`);
```
Comment (ArmanNik) [inline src/routes/(console)/organization-[organization]/change-plan/+page.svelte:220]: I wouldn't redirect always to the billing tab... We could redirect to `previousPage` so they end up where they started :+1: 

(same comment applies to the other instance we do this)

src/routes/(console)/organization-[organization]/settings/deleteOrganizationModal.svelte
```diff
@@ -55,7 +55,7 @@
             showDelete = false;
             addNotification({
                 type: 'success',
-                message: `${$organization.name} has been flagged for deletion`
+                message: `${$organization.name} has been deleted`
```
Comment (ArmanNik) [inline src/routes/(console)/organization-[organization]/settings/deleteOrganizationModal.svelte:58]: The linear task says that the copy needs to change only when a free org is deleted, the old copy still works for pro and scale orgs

ArmanNik (changes requested):

src/routes/(console)/organization-[organization]/change-plan/+page.svelte
Comment (ernstmul) [reply src/routes/(console)/organization-[organization]/change-plan/+page.svelte]: Fixed!

src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
Comment (ernstmul) [reply src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte]: Changed!

ArmanNik (approved):

PR #1426, opened 2024-10-17
feat: improve security stance

ArmanNik:


TorstenDittmann:
That approach might not work, because we are building a static site. We probably have to do it here https://github.com/appwrite/console/blob/main/docker/nginx.conf#L25-L32 with a specific config file for cloud 👍🏻

eldadfux:

PR #1487, opened 2024-11-01
Skip serviceworker cache install on oauth pages

ernstmul:
## What does this PR do?

Exclude the oauth domains from the service worker so no cache is initialised (since it's not needed)

✅

### DIFF
--- a/src/service-worker.ts
+++ b/src/service-worker.ts
@@ -23,7 +23,7 @@ sw.addEventListener('install', (event) => {
     }
 
     // Check if the page URL does not match the exclusion path
-    const shouldCache = !base.startsWith('/console/auth');
+    const shouldCache = !location.pathname.startsWith(`${base}/auth`);
     if (shouldCache) {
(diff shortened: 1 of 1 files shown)


src/service-worker.ts
```diff
@@ -22,7 +22,11 @@ sw.addEventListener('install', (event) => {
         await cache.addAll(ASSETS);
     }
 
-    event.waitUntil(addFilesToCache());
+    // Check if the page URL does not match the exclusion path
+    const shouldCache = !/\/console\/auth\/oauth2\/.*/.test(location.pathname);
```
TorstenDittmann, line 26: I think we can do `location.pathname.startswith('/console/auth')` and we can also use the `base` variable from the `$service-worker` module import from top

ernstmul: Makes sense! Updated it!

TorstenDittmann (approved):

PR #1518, opened 2024-11-15
fix: prefs empty array case

TorstenDittmann:
## What does this PR do?

- fallback check to account for an empty array being returned was not accessiing possible null values correctly

## Test Plan

- manual

✅

### DIFF
--- a/src/lib/helpers/notifications.ts
+++ b/src/lib/helpers/notifications.ts
@@ -24,7 +24,7 @@ const notificationPrefs = (): Record<string, NotificationPrefItem> => {
     }
 
     // for some reason, the prefs become array as default or on all clear. let's reset.
-    return Array.isArray(prefs.notificationPrefs) ? {} : prefs.notificationPrefs || {};
+    return Array.isArray(prefs?.notificationPrefs) ? {} : prefs?.notificationPrefs ?? {};
 };
 
 function updateNotificationPrefs(parsedPrefs: Record<string, NotificationPrefItem>) {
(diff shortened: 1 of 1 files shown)


src/lib/helpers/notifications.ts
```diff
@@ -24,7 +24,7 @@ const notificationPrefs = (): Record<string, NotificationPrefItem> => {
     }
 
     // for some reason, the prefs become array as default or on all clear. let's reset.
-    return Array.isArray(prefs.notificationPrefs) ? {} : prefs.notificationPrefs || {};
+    return Array.isArray(prefs?.notificationPrefs) ? {} : (prefs?.notificationPrefs ?? {});
```
stnguyen90, line 27: There's already a null check above, though 🧐

stnguyen90: Ahhh it's a null check above...maybe that check should also check for undefined?

src/lib/helpers/notifications.ts
```diff
@@ -19,12 +19,12 @@ const userPreferences = () => get(user)?.prefs;
 const notificationPrefs = (): Record<string, NotificationPrefItem> => {
     const prefs = userPreferences();
 
-    if (prefs === null) {
+    // due to php backend, empty object can be returnd as an empty array
+    if (prefs?.notificationPrefs === null || Array.isArray(prefs?.notificationPrefs)) {
```
stnguyen90, line 23: what if `prefs` is `undefined`?

TorstenDittmann: also fixed 👍🏻 improved the condition handling

stnguyen90 (approved):

PR #1553, opened 2024-12-12
Fix terms checkbox and gap

ItzNotABug:
## What does this PR do?

This PR addresses the gap between the title and the checkbox while also addressing the issue where if the checkbox is checked, it cannot be unchecked.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/src/lib/elements/forms/inputChoice.svelte
+++ b/src/lib/elements/forms/inputChoice.svelte
@@ -30,7 +30,7 @@
     }
 </script>
 
-<label class="choice-item" for={id}>
+<div class="choice-item">
     <Layout.Stack direction="row" alignItems="flex-start">
         {#if type === 'switchbox'}
             <Selector.Switch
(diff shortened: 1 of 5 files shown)


src/lib/elements/forms/inputChoice.svelte
```diff
@@ -30,7 +30,7 @@
     }
 </script>
 
-<label class="choice-item" for={id}>
+<div class="choice-item">
```
ernstmul, line 33: This now means that clicking the label text doesn't actually change the state of the Selector right?

ItzNotABug: @ernstmul 
It should, iirc, the click wasn't working due to nested labels. This outer label and there's one in Base under PinkV2.

ernstmul (approved):

PR #1564, opened 2024-12-16
New layout for support and add subject

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


src/lib/components/support.svelte
```diff
@@ -1,123 +1,159 @@
...
+        flex-direction: column;
+    }
+
+    @media (max-width: 768px) {
+        .support-section {
+            gap: 1.25rem;
+            padding: 0.5rem;
+        }
+    }
+</style>
```
ernstmul, line 159: Let's change this to lang='scss', and then use the nesting to not have to retype the class name.

ernstmul (approved):

src/lib/components/support.svelte
```diff
@@ -1,123 +1,162 @@
...
+
                 <p class="u-line-height-1-5 u-margin-block-start-8">
-                    Get personalized support from the Appwrite team from <b>{supportTimings}</b>
+                    {option.description}
                 </p>
+            </div>
+
+            {#if option.showSupport}
+                <div class="u-flex u-gap-12 u-cross-center">
+                    {#if !isPaid}
```
stnguyen90, line 76: instead of looking at the plan for premium support, let's look at the current org's plan and check if `premiumSupport` is true.

stnguyen90 (changes requested):

src/lib/components/support.svelte
ItzNotABug: Done.

src/lib/components/support.svelte
```diff
@@ -1,120 +1,159 @@
...
+            label: 'Community support',
+            link: 'https://appwrite.io/discord',
+            description: 'Get support from our community through Discord'
+        },
+        {
+            icon: 'github',
+            cta: 'Open issue',
+            showSupport: false,
+            label: 'Open GitHub issue',
+            link: 'https://github.com/appwrite',
```
stnguyen90, line 49: Maybe it would be better to link [here](https://github.com/appwrite/appwrite/issues/new/choose) so it's less clicks and there's less thinking involved in choosing a repo?

```suggestion
            link: 'https://github.com/appwrite/appwrite/issues/new/choose',
```

src/lib/helpers/date.ts
```diff
@@ -128,6 +128,11 @@ export const localeTimezoneName = () => {
     return dateWithTimezone.split(', ')[1];
 };
 
+export const localeShortTimezoneName = () => {
+    const timezone = localeTimezoneName();
+    return timezone.match(/[A-Z]/g)?.join('') || timezone;
+};
+
```
stnguyen90, line 135: Why can't we use `shortGeneric` for `timeZoneName` in `Intl.DateTimeFormatOptions` instead?

src/lib/components/support.svelte
```diff
@@ -1,120 +1,159 @@
...
+            description: 'Report a bug or pitch a new feature'
+        }
+    ];
 </script>
 
-{#if isCloud}
```
stnguyen90, line 20: Some things shouldn't be shown on self-hosted, right?

stnguyen90 (changes requested):

src/lib/helpers/date.ts
ItzNotABug: 1. doesn't fit

  image

2. using regex replace gives a better tz name like `IST` whereas using `shortGeneric` gives `India Time`

  image

src/lib/components/support.svelte
ItzNotABug: Done.

src/lib/components/support.svelte
```diff
@@ -72,49 +150,44 @@
...
+        background: #131315 !important;
+    }
+
+    :global(.theme-light .support-option-card) {
+        border: 1px solid var(--color-border-neutral, #ededf0);
+        background: var(--color-bgColor-neutral-default, #fafafb);
+    }
+
+    :global(.theme-light .support-option-card .secondary-button) {
+        background: #fff !important;
```
ernstmul, line 191: Can you check all the !importants. And for the hardcoded color vars. Do we have Pink v2 vars for those? We could implement them straight away, with the hardcoded value as fallback. To future proof this.

src/routes/(console)/wizard/support/mobileSupportModal.svelte
```diff
@@ -5,6 +5,24 @@
...
+</div>
+
+<style>
+    :global(.mobile-support-dialog-modal dialog) {
+        background-color: #fff;
+    }
+
+    :global(.theme-dark .mobile-support-dialog-modal dialog) {
+        background-color: #131315;
+    }
```
ernstmul, line 23: Same about pink v2 color vars with fallback

src/lib/components/support.svelte
ItzNotABug: I think we do have the pink2 vars, let me check.

src/lib/helpers/date.ts
stnguyen90: so frustrating...it looks like we should be using `short`, but it'll show as `GMT+5:30` rather than `IST` because `IST` is vague and can be Israel Standard Time or Irish Standard Time, etc.

Reference: https://github.com/tc39/proposal-temporal/issues/2257#issuecomment-1152070209

Can you add a comment explaining why we need this workaround?

stnguyen90 (approved):

PR #1574, opened 2024-12-20
Add check to which plan the org is changing when applying the credits

ernstmul:
## What does this PR do?
Applying credits can make an organisation switch plans. This additional check prevents an organisation from downgrading when applying credits.

✅

### DIFF
--- a/src/routes/(console)/+layout.ts
+++ b/src/routes/(console)/+layout.ts
@@ -25,7 +25,6 @@ export const load: LayoutLoad = async ({ fetch, depends, parent }) => {
     let plansInfo = new Map<Tier, Plan>();
     if (isCloud) {
         const plansArray = await sdk.forConsole.billing.getPlansInfo();
-        console.log('plansArray', plansArray);
         plansInfo = plansArray.plans.reduce((map, plan) => {
             map.set(plan.$id as Tier, plan);
             return map;
--- a/src/routes/(console)/+layout.ts
+++ b/src/routes/(console)/+layout.ts
@@ -25,6 +25,7 @@ export const load: LayoutLoad = async ({ fetch, depends, parent }) => {
     let plansInfo = new Map<Tier, Plan>();
     if (isCloud) {
         const plansArray = await sdk.forConsole.billing.getPlansInfo();
+        console.log('plansArray', plansArray);
         plansInfo = plansArray.plans.reduce((map, plan) => {
             map.set(plan.$id as Tier, plan);
             return map;
--- a/src/routes/(console)/apply-credit/+page.svelte
+++ b/src/routes/(console)/apply-credit/+page.svelte
@@ -190,17 +190,11 @@
         (team) => team.$id === selectedOrgId
     ) as Organization;
 
-    function getNewBillingPlan(organization: Organization): BillingPlan {
-        if (organization?.billingPlan === BillingPlan.SCALE) {
-            return BillingPlan.SCALE;
-        } else if (campaign?.plan) {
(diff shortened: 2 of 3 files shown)


src/routes/(console)/apply-credit/+page.svelte
```diff
@@ -189,6 +189,18 @@
...
+        if (organization?.billingPlan === BillingPlan.SCALE) {
+            return BillingPlan.SCALE;
+        } else if (campaign?.plan) {
+            return campaign.plan;
+        } else {
+            return BillingPlan.PRO;
+        }
+    }
+
+    $: billingPlan = getNewBillingPlan(selectedOrg);
```
stnguyen90, line 203: @ArmanNik and @TorstenDittmann might now more on Svelte best practices, but I don't think we need to create a new function for this. You can put an if statement directly here like:

https://github.com/appwrite/console/blob/4145174c81d72ec46f62b11538e3cf2392674dfc/src/lib/layout/logs.svelte#L66-L68

src/routes/(console)/apply-credit/+page.svelte
```diff
@@ -189,6 +189,18 @@
     $: selectedOrg = $organizationList?.teams?.find(
         (team) => team.$id === selectedOrgId
     ) as Organization;
+
+    function getNewBillingPlan(organization: Organization): BillingPlan {
+        if (organization?.billingPlan === BillingPlan.SCALE) {
+            return BillingPlan.SCALE;
```
stnguyen90, line 195: I'm still a bit hesitant to hardcode stuff here. For example, enterprise plans will run into a problem with this logic.

Maybe we can rely on the plan order (after it gets fixed for the education plan)?

src/routes/(console)/apply-credit/+page.svelte
ernstmul: For now I've changed it to a ternary operation. The if statement way didn't seem to do the trick correctly (perhaps @TorstenDittmann or @ArmanNik can point out later on what I'm doing wrong there haha).

src/routes/(console)/apply-credit/+page.svelte
ernstmul: Yeah definitely! However atm we don't have a plan order in code afaik?

src/routes/(console)/+layout.ts
```diff
@@ -25,6 +25,7 @@ export const load: LayoutLoad = async ({ fetch, depends, parent }) => {
     let plansInfo = new Map<Tier, Plan>();
     if (isCloud) {
         const plansArray = await sdk.forConsole.billing.getPlansInfo();
+        console.log('plansArray', plansArray);
```
ItzNotABug, line 28: leftover?

src/routes/(console)/apply-credit/+page.svelte
ItzNotABug: @ernstmul We don't have an exact order [there's `tier-*` format though] but possibly we could check via `getNextTier`, not 100% sure as doesn't include anything about enterprise/custom tiers atm.

ItzNotABug (commented):
Minor nitpick

src/routes/(console)/apply-credit/+page.svelte
ernstmul: `tier-*` doesn't hold up anymore, since we now also have auto-1 and cont-1! `getNextTier` would definitely work, but perhaps its nice get this info from BE?

ItzNotABug (approved):

PR #1575, opened 2024-12-20
Filter the new users count to not include future days

ernstmul:
## What does this PR do?

Before:
image

After:
image

Usage data comes from the API as newly added users per day. The graph is being built by subtracting these newly added users from the total amount. The API returns future dates (with 0 new users added on those day, because they are in the future). So the graph showed days in the future. This is now fixed!

✅

### DIFF
--- a/src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte
+++ b/src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte
@@ -143,9 +143,11 @@
         <Heading tag="h6" size="7">Users</Heading>
 
         <p class="text">The total number of users across all projects in your organization.</p>
-
         <svelte:fragment slot="aside">
             {#if data.organizationUsage.users}
+                {@const users = data.organizationUsage.users.filter(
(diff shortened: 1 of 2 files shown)


src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte
```diff
@@ -143,9 +143,11 @@
         <Heading tag="h6" size="7">Users</Heading>
 
         <p class="text">The total number of users across all projects in your organization.</p>
-
         <svelte:fragment slot="aside">
             {#if data.organizationUsage.users}
+                {@const users = data.organizationUsage.users.filter(
+                    (user) => new Date(user.date) < new Date()
+                )}
```
Comment (ItzNotABug) [inline src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte:150]: Would it make more sense to move this on the `.ts` side? I see bunch of calc already being handled there. wdyt?

Comment (ernstmul) [reply src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte:150]: Yeah would also make more sense typing wise. I'll look into it!

ItzNotABug (approved):

PR #1587, opened 2024-12-30
Fix pagination

ItzNotABug:
## What does this PR do?

Fixes pagination on console when using Pink V2.

## Test Plan

Manual.

Before -

image

After -

image

## Related PRs and Issues

N/A.

Yes.


ernstmul:
Imo I would say remove all the commented stuff. If we find out we do have a need for it after Pink v2 we always have it in GIT versioning. Otherwise this file becomes like 80% commented out code, which we'll probably otherwise never really remove.

ItzNotABug:
> Imo I would say remove all the commented stuff. If we find out we do have a need for it after Pink v2 we always have it in GIT versioning. Otherwise this file becomes like 80% commented out code, which we'll probably otherwise never really remove.

Makes sense. Fixing rq.

ernstmul (approved):

PR #1601, opened 2025-01-08
Start building PRO flow

ernstmul:
## What does this PR do?

When visiting the console through the "Start building" button on the PRO column of the site pricing, you get redirected to the console.

**when not signed in**,
- the console redirects you to register
- after registration you get redirected to register an organisation with PRO plan preselected

**when signed in**,
- the console redirects you to register an organisation with PRO plan preselected

✅

### DIFF
--- a/src/lib/helpers/pricingRedirect.ts
+++ b/src/lib/helpers/pricingRedirect.ts
@@ -1,0 +1,13 @@
+import { goto } from '$app/navigation';
+import { base } from '$app/paths';
+
+export function checkPricingRefAndRedirect(searchParams: URLSearchParams, shouldRegister = false) {
+    if (searchParams.has('type')) {
+        const paramType = searchParams.get('type');
+        if (paramType === 'createPro') {
+            shouldRegister
+                ? goto(`${base}/register?type=createPro`)
+                : goto(`${base}/create-organization?type=createPro`);
+        }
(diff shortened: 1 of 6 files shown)


src/lib/helpers/pricingRedirect.ts
```diff
@@ -0,0 +1,13 @@
...
+import { base } from '$app/paths';
+
+export function checkPricingRefAndRedirect(searchParams: URLSearchParams, shouldRegister = false) {
+    if (searchParams.has('type')) {
+        const paramType = searchParams.get('type');
+        if (paramType === 'createPro') {
+            shouldRegister
+                ? goto(`${base}/register?type=createPro`)
+                : goto(`${base}/create-organization?type=createPro`);
+        }
```
ItzNotABug, line 11: do we know if we are going to open scale as well, like `createScale` in the params? I am not yet aware of the Scale GA. If not, lgtm.

ernstmul: Currently on the website that button is still disabled, and also in the console it's not a visible option. So I thought: we'll cross that bridge when we get there. It is the reason I moved everything into 1 function, so we can easily adjust it when we do need to make that change!

ItzNotABug (approved):

PR #1618, opened 2025-01-20
Show only the size limit for the bucket

ernstmul:
## What does this PR do?

Alert about file limit is only shown when file actually exceeds the limit. Also the size shown is now the bucket upload limit, with a reference to the bucket settings.
image

✅

### DIFF
--- a/src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte
+++ b/src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte
@@ -1,43 +1,37 @@
 <script lang="ts">
     import { Alert, CustomId } from '$lib/components';
-    import { BillingPlan } from '$lib/constants';
+    import { page } from '$app/stores';
     import { Pill } from '$lib/elements';
-    import { Button, FormList, InputFile } from '$lib/elements/forms';
+    import { FormList, InputFile } from '$lib/elements/forms';
     import { humanFileSize, sizeToBytes } from '$lib/helpers/sizeConvertion';
     import WizardStep from '$lib/layout/wizardStep.svelte';
-    import { tierToPlan, upgradeURL } from '$lib/stores/billing';
-    import { organization, currentPlan } from '$lib/stores/organization';
+    import { currentPlan } from '$lib/stores/organization';
     import { isCloud } from '$lib/system';
     import { bucket } from '../store';
     import { createFile } from './store';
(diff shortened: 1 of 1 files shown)


src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte
```diff
@@ -1,43 +1,37 @@
...
         <Alert type="info">
             <p class="text">
-                The {plan.name} plan has a maximum upload file size limit of {Math.floor(
-                    parseInt(size.value)
-                )}{size.unit}.
-                {#if $organization?.billingPlan === BillingPlan.FREE}
-                    Upgrade to allow files of a larger size.
-                {/if}
+                The maximum file upload size for this bucket is {parseInt(size.value)}{size.unit}.
+                Edit this number in your
```
Comment (ItzNotABug) [inline src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte:29]: lets confirm this copy with design.

src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte
```diff
@@ -47,6 +41,7 @@
                 required
                 bind:files={$createFile.files}
                 allowedFileExtensions={$bucket.allowedFileExtensions}
+                bind:error={fileError}
                 maxSize={isCloud
                     ? $bucket.maximumFileSize
                     : ($bucket.maximumFileSize ?? sizeToBytes(service, 'MB', 1000))} />
```
Comment (ItzNotABug) [inline src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte:47]: is it possible to somehow re-use the `size` const declared above, I see similar logic for specifying the max size.

ItzNotABug (changes requested):

src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte
Comment (ernstmul) [reply src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte:29]: This is 1:1 copy from the Figma :)

src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte
Comment (ernstmul) [reply src/routes/(console)/project-[project]/storage/bucket-[bucket]/create-file/step1.svelte:47]: The logic of how big a file is that is ready to be uploaded is all in the `InputFile`, so I wanted to tap into the logic that is already there instead of redoing it in the parent component.

PR #1652, opened 2025-02-08
Fix invalid hostnames on platform creation

ItzNotABug:
## What does this PR do?

Fixes an issue where invalid hostnames were allowed for web platforms.

## Test Plan

Manual.

## Related PRs and Issues

* https://github.com/appwrite/appwrite/issues/9293

Yes.

### DIFF
--- a/src/lib/layout/wizard.svelte
+++ b/src/lib/layout/wizard.svelte
@@ -68,10 +68,12 @@
     }
 
     async function submit() {
+        let interceptorResponse = null;
+
         if ($wizard.interceptor) {
             $wizard.nextDisabled = true;
             try {
-                await $wizard.interceptor();
+                interceptorResponse = await $wizard.interceptor();
             } catch (error) {
                 if (!$wizard.interceptorNotificationEnabled) return;
                 addNotification({
@@ -83,6 +85,10 @@
                 $wizard.nextDisabled = false;
             }
         }
+
+        // maintains backward compatibility.
+        // explicit boolean check as the return type can be undefined/void too!
+        if (typeof interceptorResponse === 'boolean' && interceptorResponse === false) return;
 
         wizard.setInterceptor(null);
         if (isLastStep) {
--- a/src/lib/layout/wizard.svelte
+++ b/src/lib/layout/wizard.svelte
@@ -68,12 +68,10 @@
     }
 
     async function submit() {
-        let interceptorResponse = null;
-
         if ($wizard.interceptor) {
             $wizard.nextDisabled = true;
             try {
-                interceptorResponse = await $wizard.interceptor();
+                await $wizard.interceptor();
             } catch (error) {
                 if (!$wizard.interceptorNotificationEnabled) return;
                 addNotification({
@@ -85,10 +83,6 @@
                 $wizard.nextDisabled = false;
             }
         }
-
-        // maintains backward compatibility.
-        // explicit boolean check as the return type can be undefined/void too!
-        if (typeof interceptorResponse === 'boolean' && interceptorResponse === false) return;
 
         wizard.setInterceptor(null);
         if (isLastStep) {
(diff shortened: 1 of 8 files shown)


src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte
```diff
@@ -84,6 +85,11 @@
     };
 
     async function beforeSubmit() {
+        // double-check the hostname value.
+        if (platform === PlatformType.Flutterweb && !isHostnameValid($createPlatform.hostname)) {
+            throw new Error('Please enter a valid hostname');
+        }
+
```
Comment (stnguyen90) [inline src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte:92]: I thought we typically pass a regex to the input for these kinds of validations. Why can't we do that here?

Comment (ItzNotABug) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte:92]: 1. extracted function can be helpful for future
2. We could use regex pattern, but this way also helps in disabling the next button

ernstmul (approved):

src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte
Comment (stnguyen90) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte:92]: Re-usable is good.

I still have some concerns:
1. This pattern is different compared to other places in the Console.
2. There is little feedback to the user about what's wrong.

For example, here:

image

Yes, the button is disabled, but I don't know why. Compared to other places like create document: 

image

As you can see for the `asdf` attribute, there's more feedback to the developer about what's wrong.

Comment (ItzNotABug) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte:92]: Makes sense. I can add a regex with the existing logic and that should give a clearer idea on why its disabled. Thanks for the screenshots 👍

Comment (ItzNotABug) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte]: @stnguyen90 made a few changes to overall logic. It doesn't seem possible to directly manage the warning helper via the `InputText` due to regex complexity as Svelte breaks it. Let me know if the newer changes are OK (backwards compatible).

Comment (stnguyen90) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte]: Ya, we've dealt the regex problem in the past. It seems there's a special type of regex that's used in HTML inputs

src/lib/layout/wizard.svelte
```diff
@@ -84,6 +86,10 @@
             }
         }
 
+        // maintains backward compatibility.
+        // explicit boolean check as the return type can be undefined/void too!
+        if (typeof interceptorResponse === 'boolean' && interceptorResponse === false) return;
```
stnguyen90, line 91: does this feel a little hacky?

src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte
Comment (stnguyen90) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte]: What if we update the inputText input to expose a pattern and patternError like inputOTP:

https://github.com/appwrite/console/blob/e5ab42659fdc5324e2fc3734481c8c48f2c7b985/src/lib/elements/forms/inputOTP.svelte#L12-L13

As for the pattern, I think this should work:

```
const pattern = String.raw`(\*\.)?(?!-)[A-Za-z0-9\-]+([\-\.]{1}[a-z0-9]+)*\.[A-Za-z]{2,18}`;
```

We might also want to create a new input because we're going to also need it on the update platform pages such as:

https://github.com/appwrite/console/blob/e5ab42659fdc5324e2fc3734481c8c48f2c7b985/src/routes/(console)/project-%5Bproject%5D/overview/platforms/%5Bplatform%5D/web.svelte#L48-L53

Comment (ItzNotABug) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte]: That regex wouldn't have allowed `*`, updated with latest changes.  No changes to wizard required now.

Comment (stnguyen90) [reply src/routes/(console)/project-[project]/overview/platforms/wizard/flutter/step1.svelte]: Good catch!

PR #1664, opened 2025-02-13
Feat: routes for viewing invoice

lohanidamodar:
## What does this PR do?

- Add route to view PDF invoice
- Add route to download PDF invoice
- Each route validates that invoice exists and then redirects to the REST endpoint

## Test Plan

- Manually tested it redirects to correct endpoint upon success
- Manually tested it throws error if invoice doesn't exist

## Related PRs and Issues

YES

### DIFF
--- a/src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
+++ b/src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
@@ -42,7 +42,6 @@
     };
 
     const limit = 5;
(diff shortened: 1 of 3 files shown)


ItzNotABug (approved):
we already handle navigating to a different URL if the visitor doesn't have a session, right?

lohanidamodar:
> we already handle navigating to a different URL if the visitor doesn't have a session, right?

Yes we do.

PR #1666, opened 2025-02-13
Fix: Usage alert

ItzNotABug:
## What does this PR do?

Hides the `You've reached the limit for the {PLAN} plan` if an invoice has failed.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/src/lib/layout/containerHeader.svelte
+++ b/src/lib/layout/containerHeader.svelte
@@ -86,12 +86,14 @@
 
 <!-- Show only if on Cloud, alerts are enabled, and it isn't a project limited service -->
 {#if isCloud && showAlert}
-    {#if $readOnly}
-        {@const services = overflowingServices
-            .map((s) => {
-                return s.name.toLocaleLowerCase();
-            })
-            .join(', ')}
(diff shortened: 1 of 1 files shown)


src/lib/layout/containerHeader.svelte
```diff
@@ -86,12 +86,14 @@
...
-            })
-            .join(', ')}
+    <!-- some services are above limit -->
+    {@const services = overflowingServices
+        .map((s) => {
+            return s.name.toLocaleLowerCase();
+        })
+        .join(', ')}
+
+    {#if $readOnly && services.length}
```
stnguyen90, line 96: This might still be a problem because `$readOnly` is only true when there are failed invoices, right?

Also, does this mean some of these things don't show if they don't have access (non owner role in the org)?!

lohanidamodar (approved):

src/lib/layout/containerHeader.svelte
stnguyen90: > Yea, seems we should remove readOnly check as we do overflowing services check.

Ya makes sense.

PR #1679, opened 2025-02-19
fix: show multi-region cname target.

ItzNotABug:
## What does this PR do?

Show correct domain for cname mapping.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/src/lib/pages/domains/wizard/cnameTable.svelte
+++ b/src/lib/pages/domains/wizard/cnameTable.svelte
@@ -11,11 +11,16 @@
         TableRow
     } from '$lib/elements/table';
     import { domain } from './store';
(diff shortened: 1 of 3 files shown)


src/lib/pages/domains/wizard/cnameTable.svelte
```diff
@@ -27,6 +32,7 @@
     </TableHeader>
     <TableBody>
         <TableRow>
+            {@const target = targetValue()}
```
ernstmul, line 35: Why don't we use `targetValue()` straight away instead of assigning it to a var first?

ernstmul (approved):

PR #1683, opened 2025-02-19
fix: apply credit flow

ArmanNik:


static/images/campaigns/general_50/dark.png
TorstenDittmann, line 1: general question, didnt we remove all campaign images from the console? 🤔

src/routes/+layout.svelte
```diff
@@ -47,7 +47,12 @@
                     .getCampaign(coupon.campaign)
                     .catch<null>(() => null);
                 if (campaign && $user) {
-                    goto(`${base}/apply-credit?code=${code}`);
+                    const params = new URLSearchParams($page.url.searchParams);
+                    params.delete('code'); // Remove the code param since we're using it in the new path
+                    const searchString = params.toString();
+                    goto(
+                        `${base}/apply-credit?code=${code}${searchString ? '&' + searchString : ''}`
+                    );
```
TorstenDittmann, line 55: why do we remove it and add the `code` param again? :-D

TorstenDittmann: shouldn't 

```js
                    goto(
                        `${base}/apply-credit?${$page.url.searchParams}`
                    )
```
do the same thing?

static/images/campaigns/general_50/dark.png
ArmanNik: All images are still in the console, I think we just get a path from backend that points to the related images

TorstenDittmann (approved):

PR #1695, opened 2025-02-22
Update message status via polling

ItzNotABug:
## What does this PR do?

Messaging doesn't currently have realtime. Due to this the status is not updated when its sent in the details screen.
This PR fixes that by adding a polling mechanism to messages.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/src/routes/(console)/project-[project]/messaging/helper.ts
+++ b/src/routes/(console)/project-[project]/messaging/helper.ts
@@ -31,7 +31,7 @@ export function stopPolling() {
     messageIntervals.forEach((_, messageId) => clearPolling(messageId));
 }
 
-/** Starts polling message statuses, optionally marking them as processing. */
+/** Starts polling message statuses. */
 export function pollMessagesStatus(messages: Models.Message[]) {
     messages.forEach((message) => {
         clearPolling(message.$id);
--- a/src/routes/(console)/project-[project]/messaging/helper.ts
+++ b/src/routes/(console)/project-[project]/messaging/helper.ts
@@ -36,12 +36,8 @@ export function stopPolling() {
 }
 
 /** Starts polling message statuses, optionally marking them as processing. */
-export function pollMessagesStatus(messages: Models.Message[], update: boolean = true) {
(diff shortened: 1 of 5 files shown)


src/routes/(console)/project-[project]/messaging/helper.ts
Comment (stnguyen90) [inline src/routes/(console)/project-[project]/messaging/helper.ts:1]: Can we reorganize this in a way to avoid creating a helper file with functions?

Comment (ItzNotABug) [reply src/routes/(console)/project-[project]/messaging/helper.ts:1]: Hmm, not sure if we can do it without adding some amount of code duplicacy around both usages of it. Let me see what we can do.

src/routes/(console)/project-[project]/messaging/helper.ts
```diff
@@ -0,0 +1,43 @@
+import { sdk } from '$lib/stores/sdk';
+import { type Models } from '@appwrite.io/console';
+
+/** Stores active polling intervals for messages. */
+const messageIntervals = new Map<string, ReturnType<typeof setInterval>>();
+
+/** Checks the status of a message and stops polling if it's no longer processing. */
+function checkMessageStatus(message: Models.Message) {
```
Comment (stnguyen90) [inline src/routes/(console)/project-[project]/messaging/helper.ts:8]: Any performance problems if there's a full page of processing messages? Also, what if the status doesn't change from processing?

Comment (ItzNotABug) [reply src/routes/(console)/project-[project]/messaging/helper.ts:8]: 1. Shouldn't be too much of an overhead. Once the page is moved from, the polling stops via `onDestroy`.
2. That's a tricky one, we could add some sort of fallback in retries or a timer but I haven't seen a case where a message was stuck on `processing` indefinitely. It fails pretty quick on an issue.

Comment (stnguyen90) [reply src/routes/(console)/project-[project]/messaging/helper.ts:8]: > I haven't seen a case where a message was stuck on processing indefinitely. It fails pretty quick on an issue.

Maybe they were more frequent in the past and we've fixed them by now.

stnguyen90 (approved):

PR #1751, opened 2025-03-17
Fix feedback create web flow

ernstmul:
## What does this PR do?
Improvements for showing correct start and env vars

✅

### DIFF
--- a/src/lib/layout/wizard.svelte
+++ b/src/lib/layout/wizard.svelte
@@ -69,6 +69,7 @@
             if (confirmExit) {
                 showExitModal = true;
             } else {
+                goto(href);
                 trackEvent('wizard_exit', {
                     from: 'button'
                 });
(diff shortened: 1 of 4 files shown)


src/lib/layout/wizard.svelte
```diff
@@ -72,6 +72,7 @@
         if (confirmExit) {
             showExitModal = true;
         } else {
+             goto(href);
```
ItzNotABug, line 75: you think the `onExit` should be fired here? 👀
```ts
if (onExit) {
  onExit();

  // clear exit
  onExit = null;
}
```

ItzNotABug (approved):

src/lib/layout/wizard.svelte
ernstmul: That part is for the confirm exit modal, not the exit of the wizard!

PR #1769, opened 2025-03-28
feat: svelte 5

TorstenDittmann:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


ArmanNik (approved):

src/hooks.client.ts
```diff
@@ -2,6 +2,7 @@ import * as Sentry from '@sentry/sveltekit';
 import { AppwriteException } from '@appwrite.io/console';
 import type { HandleClientError } from '@sveltejs/kit';
 import { isCloud, isProd } from '$lib/system';
+import { dev } from '$app/environment';
```
ItzNotABug, line 5: don't think this is used here.

src/lib/components/paginationInline.svelte
```diff
@@ -17,18 +17,15 @@
     function handleOptionClick(page: number) {
         if (currentPage !== page) {
             offset = limit * (page - 1);
-            currentPage = page;
```
ItzNotABug, line 20: was removed, but default reactive logic is different, right?

src/lib/components/filters/quickFilters.svelte
```diff
@@ -58,7 +58,6 @@
                                 $columns,
                                 analyticsSource
                             );
-                            subSheets = subSheets;
```
ItzNotABug, line 61: does this removal of reassignment affect ui changes due to missed reactivity now?

ItzNotABug (changes requested):

src/lib/components/paginationInline.svelte
TorstenDittmann: it was never needed because of the `$:` syntax. the new linter complains about this

src/hooks.client.ts
TorstenDittmann, line 5:
```suggestion
```

TorstenDittmann: good catch

src/lib/components/paginationInline.svelte
ItzNotABug: yeah, but the logic on `$:` and manual assignment is diff no.? that yields same result?

TorstenDittmann: it actually triggers it twice 👍🏻  one change with changing offset and then the re-assignment

PR #1777, opened 2025-04-03
Fix: change-plan from `account` route

ItzNotABug:
## What does this PR do?

Change plan route would crash if navigated through Account > Support > Get premium support because there's no organization store or info. available. This PR fixes that.

## Test Plan

Manual.

https://github.com/user-attachments/assets/33ed36d7-d129-4152-8c67-51eb355479b5

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/src/lib/components/support.svelte
+++ b/src/lib/components/support.svelte
@@ -5,17 +5,29 @@
     import { isSupportOnline, showSupportModal } from '$routes/(console)/wizard/support/store';
     import { trackEvent } from '$lib/actions/analytics';
     import { localeShortTimezoneName, utcHourToLocaleHour } from '$lib/helpers/date';
-    import { upgradeURL } from '$lib/stores/billing';
+    import { plansInfo } from '$lib/stores/billing';
     import { Card } from '$lib/components/index';
     import { app } from '$lib/stores/app';
-    import { currentPlan } from '$lib/stores/organization';
+    import { currentPlan, organizationList } from '$lib/stores/organization';
     import { isCloud } from '$lib/system';
+    import { base } from '$app/paths';
 
     export let show = false;
 
     export let showHeader = true;
 
-    $: hasPremiumSupport = $currentPlan?.premiumSupport ?? false;
+    $: hasPremiumSupport = $currentPlan?.premiumSupport ?? areAllOrganizationsPaid ?? false;
+
+    $: areAllOrganizationsPaid = $organizationList.teams.every(
+        (team) => $plansInfo.get(team['billingPlan'])?.premiumSupport
+    );
+
+    // there can only be one free organization
+    $: freeOrganization = $organizationList.teams.find(
+        (team) => !$plansInfo.get(team['billingPlan'])?.premiumSupport
+    );
+
+    $: upgradeURL = `${base}/organization-${freeOrganization.$id}/change-plan`;
 
     $: supportTimings = `${utcHourToLocaleHour('16:00')} - ${utcHourToLocaleHour('00:00')} ${localeShortTimezoneName()}`;
 
@@ -54,7 +66,7 @@
         }
     ];
 
-    const showCloudSupport = (index) => {
+    const showCloudSupport = (index: number) => {
         return (index === 0 && isCloud) || index > 0;
     };
 </script>
@@ -82,7 +94,7 @@
                     <div class="u-flex u-gap-12 u-cross-center">
                         {#if !hasPremiumSupport}
                             <Button
-                                href={$upgradeURL}
+                                href={upgradeURL}
                                 on:click={() => {
                                     trackEvent('click_organization_upgrade', {
                                         from: 'button',
--- a/src/routes/(console)/organization-[organization]/change-plan/+page.svelte
+++ b/src/routes/(console)/organization-[organization]/change-plan/+page.svelte
@@ -28,13 +28,13 @@
         WizardSecondaryFooter
     } from '$lib/layout';
     import { type Coupon, type PaymentList } from '$lib/sdk/billing';
-    import { isOrganization, plansInfo, tierToPlan, type Tier } from '$lib/stores/billing';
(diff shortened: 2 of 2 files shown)


src/routes/(console)/organization-[organization]/change-plan/+page.svelte
```diff
@@ -274,9 +273,15 @@
...
-    $: isDowngrade = isSamePlan ? false : $plansInfo.get(billingPlan)?.order < $currentPlan.order;
+    $: selectedOrg =
+        $organization ??
+        ($organizationList.teams.find(
+            (team) => team.$id === $page.params.organization
+        ) as Organization);
+    $: currentOrgPlan = $currentPlan ?? $plansInfo.get(selectedOrg?.billingPlan);
+    $: isSamePlan = billingPlan === currentOrgPlan.$id;
+    $: isUpgrade = isSamePlan ? false : $plansInfo.get(billingPlan)?.order > currentOrgPlan.order;
+    $: isDowngrade = isSamePlan ? false : $plansInfo.get(billingPlan)?.order < currentOrgPlan.order;
```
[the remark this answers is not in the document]
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/change-plan/+page.svelte:284]: yep. planned for the billing's major refactor.

src/lib/components/support.svelte
```diff
@@ -5,17 +5,29 @@
...
+    import { base } from '$app/paths';
 
     export let show = false;
 
     export let showHeader = true;
 
-    $: hasPremiumSupport = $currentPlan?.premiumSupport ?? false;
+    $: hasPremiumSupport = $currentPlan?.premiumSupport ?? areAllOrganizationsPaid ?? false;
+
+    $: areAllOrganizationsPaid = $organizationList.teams.every(
```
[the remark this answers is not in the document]
ItzNotABug: yep, will update.

src/lib/components/support.svelte
```diff
@@ -5,17 +5,29 @@
...
+    $: allOrgsHavePremiumSupport = $organizationList.teams.every(
+        (team) => $plansInfo.get((team as Organization).billingPlan)?.premiumSupport
+    );
+
+    // there can only be one free organization
+    $: freeOrganization = $organizationList.teams.find(
+        (team) => !$plansInfo.get((team as Organization).billingPlan)?.premiumSupport
+    );
+
+    $: upgradeURL = `${base}/organization-${freeOrganization.$id}/change-plan`;
```
stnguyen90, line 30: do we need the one in the billing store if we have this here? Maybe it would be best to consolidate?

src/lib/components/support.svelte
```diff
@@ -5,17 +5,29 @@
...
     export let show = false;
 
     export let showHeader = true;
 
-    $: hasPremiumSupport = $currentPlan?.premiumSupport ?? false;
+    $: hasPremiumSupport = $currentPlan?.premiumSupport ?? allOrgsHavePremiumSupport ?? false;
+
+    $: allOrgsHavePremiumSupport = $organizationList.teams.every(
+        (team) => $plansInfo.get((team as Organization).billingPlan)?.premiumSupport
+    );
```
stnguyen90, line 23: I'd recommend allowing premiums support if they have at least 1 org with premium support.

src/lib/components/support.svelte
ItzNotABug: Yep, we still need the other one because it uses the `organization` available from either page or the derived store. On `account` route, there's no `organization` access available. We could make this a function that has a default value but that goes outside the scope of this one and would require a refactor in some other places too.

src/lib/components/support.svelte
ItzNotABug: We actually cannot/should not.

The issue is that we have a very different flow on Project > Support & Account > Support.
1. On project or Org routes, we know what to check for premium support checks
2. On others where Org can be multiple and no access to them, we don't have a way to use that org for premium support because an issue could be tied to some other Project's org.

> I'd recommend allowing premiums support if they have at least 1 org with premium support.

But then, if they have an issue with another organization, those projects won't show in the dropdown.

---

I'd rather suggest something like this if we don't want this -
1. Have Project or Org access > Show normal flow > Premium support or Get premium support
2. No access to Project or Org. > Show an Org selector dropdown > Based on selected Org, check if they have premium support > Support OR Get premium support.

WDYT?

stnguyen90: > On others where Org can be multiple and *no access to them*

What do you mean? Don't we have the list of the user's orgs here:

https://github.com/appwrite/console/blob/4e978bc7f1564612132e77af216d367ba73119c0/src/routes/%2Blayout.ts#L33-L36

Can't we filter that to check if there are any orgs that have support enabled?

Also, right now, the project list just fetches the 1st 25 projects irrespective of the org:

https://github.com/appwrite/console/blob/4e978bc7f1564612132e77af216d367ba73119c0/src/routes/(console)/wizard/support/step1.svelte#L15-L19

I plan on changing the support form anyway so I'll take care of adding an org dropdown and filtering by project.

ItzNotABug: > What do you mean? Don't we have the list of the user's orgs here:
We do, but assume that a user has an issue with a Project XYZ in Org 1, but premium is only on Org 2. If we allow premium support, projects from Org 1 won't show in the dropdown, right?

stnguyen90: > We do, but assume that a user has an issue with a Project XYZ in Org 1, but premium is only on Org 2. If we allow premium support, projects from Org 1 won't show in the dropdown, right?

@ItzNotABug, that's not how it works today. I can tweak it later to filter the projects. It's fine if they submit a ticket for a project they don't have the pro plan for. We'll address it in the support ticket.

stnguyen90 (approved):

src/lib/components/support.svelte
stnguyen90: At least this doesn't throw an error. I'll tweak it later to open the support window if at least 1 org has support.

PR #1782, opened 2025-04-06
Fix: search x debounce

ItzNotABug:
## What does this PR do?

Fixes an issue where the letters are cut off while typing, resulting an invalid or half-baked search string.

## Test Plan

Manual.

https://github.com/user-attachments/assets/2d0864b0-33ab-489d-b1b9-2a0fe381bfcb

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/src/lib/components/searchQuery.svelte
+++ b/src/lib/components/searchQuery.svelte
@@ -1,9 +1,10 @@
 <script lang="ts">
-    import { goto } from '$app/navigation';
-    import { page } from '$app/stores';
-    import { trackEvent } from '$lib/actions/analytics';
     import { onMount } from 'svelte';
     import { onDestroy } from 'svelte';
+    import { page } from '$app/stores';
+    import { goto } from '$app/navigation';
+    import { trackEvent } from '$lib/actions/analytics';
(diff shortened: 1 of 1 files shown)


src/lib/components/searchQuery.svelte
```diff
@@ -15,47 +16,44 @@
...
     });
 
-    onDestroy(() => {
-        search = '';
-        if (timer) {
-            clearTimeout(timer);
-        }
-    });
+    const runSearch = createDebounce((value: string) => {
+        const trimmed = value.trim();
```
ernstmul, line 28: I get why we trim. But we also do `url.searchParams.set('search', trimmed)` on line 41. So if we search something with a space, that goes right the first time. But since it's also set to the query param, and when you reload the page it goes bust because the space is missing right? Or am I missing something?

ItzNotABug: Don't spaces anyway get changed to `+`? 👀

ernstmul: True, sounds good!

ernstmul (approved):

PR #1877, opened 2025-05-22
Fix/fix relationship set and view

ernstmul:
## What does this PR do?

Fix the pre select of the option in the relation overview of the document

✅

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/table.svelte
+++ b/src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/table.svelte
@@ -199,7 +199,7 @@
                 {@const attr = $attributes.find((n) => n.key === id)}
                 <Table.Cell column={id} {root}>
                     {#if isRelationship(attr)}
-                        {@const args = displayNames?.[attr.relatedCollection] ?? ['name']}
+                        {@const args = displayNames?.[attr.relatedCollection] ?? ['$id']}
                         {#if !isRelationshipToMany(attr)}
                             {#if document[id]}
                                 {@const related = document[id]}


src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/attributes/relationship.svelte
```diff
@@ -110,7 +109,7 @@
     });
 
     // Reactive statements
-    $: getCollections(search).then((res) => (collectionList = res));
+    $: getCollections().then((res) => (collectionList = res));
```
Comment (ItzNotABug) [inline src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/attributes/relationship.svelte:112]: I don't remember as to why was this reactive.

src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/attributes/relationship.svelte
```diff
@@ -84,16 +84,15 @@
...
 
     // Lifecycle hooks
-    async function getCollections(search: string = null) {
-        const queries = search ? [Query.orderDesc('')] : [Query.limit(100)];
+    async function getCollections() {
+        const queries = [Query.limit(100)];
         return sdk
             .forProject(page.params.region, page.params.project)
-            .databases.listCollections(databaseId, queries, search);
+            .databases.listCollections(databaseId, queries);
```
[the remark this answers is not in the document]
Comment (ernstmul) [reply src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/attributes/relationship.svelte:95]: This is getting the collections from the API. The combobox is filtering on the client based on what the API returned. ATM search is defaulting as `null` which results in 0 results coming from the API. So ATM it's entirely broken, and this fixes that.
Updating this to the ComboBox is still something we need to do, but there is a bug that the dropdown of the combobox falls behind the modal, so you cannot use it.

TLDR: this change is going from 100% broken to usable. We need to do more work to make it good.

src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/attributes/relationship.svelte
Comment (ernstmul) [reply src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/attributes/relationship.svelte:112]: Probably because of the search, but with that being going its not needed anymore...

ArmanNik (approved):

PR #1884, opened 2025-05-24
Fix: hardcoded max file size values

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte
+++ b/src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte
@@ -8,7 +8,7 @@
     import { addNotification } from '$lib/stores/notifications';
     import { sdk } from '$lib/stores/sdk';
     import { Icon, Layout, Tooltip, Typography, Upload } from '@appwrite.io/pink-svelte';
-    import { writable } from 'svelte/store';
+    import { get, writable } from 'svelte/store';
     import { ID, Runtime } from '@appwrite.io/console';
     import type { Models } from '@appwrite.io/console';
     import { consoleVariables } from '$routes/(console)/store';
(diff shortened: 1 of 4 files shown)


src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte
```diff
@@ -124,6 +124,7 @@
         on:invalid={handleInvalid}
         extensions={['json']}
         bind:files={files[input.name]}
+        maxSize={10000000}
```
Comment (stnguyen90) [inline src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte:127]: Should this be hardcore?

src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte
```diff
@@ -153,6 +160,15 @@
               };
           })
         : [];
+
+    $: maxSize =
+        isCloud && $currentPlan
+            ? $currentPlan?.deploymentSize === 0
+                ? 0
```
Comment (stnguyen90) [inline src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: Why/when would this be 0? Do we even need this if because it deployment size was 0, 0 * 1000000 is also 0.

Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: for Unlimited, it is 0 on the BE based on a given plan.

src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte
Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte:127]: The text value for this was 10MB, I don't see why this shouldn't be 10MB for a JSON file. I also don't think the BE has any flag for this. Is there?

src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte
Comment (stnguyen90) [reply src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: Oh I didn't realize we had a plan with unlimited.

Anyways, in either case for the if, if deployment size was 0, the result would be 0 wouldn't it?

Comment (stnguyen90) [reply src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: So what other case is this if supposed to cover?

src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte
Comment (stnguyen90) [reply src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte:127]: Oh this is the firebase service json! I think the console actually reads the file and sends JSON.

src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte
Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: If not `0`, we still need to `* 1000000` because the value from BE is not a full sized byte value, just simpler integer like `30`/`60`, etc..

stnguyen90 (commented):
Are tests expected to fail?

ItzNotABug:
> Are tests expected to fail?

atm, yes. but here's the fix - https://github.com/appwrite/console/pull/1885

src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte
Comment (stnguyen90) [reply src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: But if deployment size is 0, 0 * 1000000 is 0, so why do we need the if?

Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: aah that way. makes sense. will fix.

Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/functions/create-function/manual/+page.svelte:167]: yepp.

stnguyen90 (approved):

PR #1904, opened 2025-05-27
fix: org routing, project creation

ArmanNik:


src/routes/(console)/organization-[organization]/createProjectCloud.svelte
```diff
@@ -51,21 +54,14 @@
...
 
     onDestroy(() => {
-        $createProject = {
-            id: null,
-            name: null,
-            region: 'fra'
-        };
+        id = null;
+        name = null;
+        region = 'fra';
```
Comment (ItzNotABug) [inline src/routes/(console)/organization-[organization]/createProjectCloud.svelte:59]: why a value after destroyed?

PR #1917, opened 2025-05-27
fix: fix invoice view and download links

stnguyen90:
## What does this PR do?

The prefix should be the api endpoint instead of the base path so that the developer goes straight to the API endpoint to view or download their invoice.

## Test Plan

Manually tested locally

## Related PRs and Issues

None

Yes


lohanidamodar (commented):
The URL is still wrong here `orgainzations/$organizationId` should be `organization-$organizationId`
Fixed here
https://github.com/appwrite/console/pull/1920

ItzNotABug (approved):

PR #1944, opened 2025-05-30
fix: credits wrong redirection

ArmanNik:
The account was invalidated too soon during registration, leading to weird behaviour


src/routes/(public)/(guest)/register/+page.svelte
```diff
@@ -32,22 +32,18 @@
...
                 return;
-            }
-            if (data?.campaign) {
+            } else if (data?.campaign?.$id) {
                 await goto(`${base}/apply-credit?campaign=${data.campaign.$id}`);
                 return;
-            }
-            if (page.url.searchParams) {
+            } else if ($redirectTo) {
+                window.location.href = $redirectTo;
```
ItzNotABug, line 44: why not `goto`? is the url external?

ItzNotABug (approved):

PR #1946, opened 2025-05-30
Always redirect to create-org after onboarding is completed once

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/src/routes/(console)/onboarding/create-project/+page.svelte
+++ b/src/routes/(console)/onboarding/create-project/+page.svelte
@@ -22,7 +22,7 @@
 
     export let data;
 
-    async function markOnboardingComplete() {
+    function markOnboardingComplete() {
         const currentPrefs = data.accountPrefs ?? $user.prefs;
 
         const newPrefs = {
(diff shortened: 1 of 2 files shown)


src/routes/(console)/onboarding/create-project/+page.svelte
```diff
@@ -31,6 +44,9 @@
                 teamId,
                 isCloud ? region : undefined
             );
+
+            await markOnboardingComplete();
```
TorstenDittmann, line 48: doe we need to await it?

ItzNotABug: maybe, maybe not.
I don't want to really wait for an edge case to happen where this fails for some reason and the next hit to onboarding route creates another org, if the last was deleted. wdyt?

TorstenDittmann: but this should never make project creation fail though 👍🏻

ItzNotABug: maybe move it to `finally`, while being non await?

TorstenDittmann: then it doesnt make a difference anyway if its awaited :-P just makes everything slower

PR #1947, opened 2025-05-30
fix: estimation box component

ArmanNik:
- Fix copy
- Use `estimation` endpoint to calculate values
- Delete duplicate component

Turns out we had two components doing the same thing, and both were wrong


TorstenDittmann (approved):

src/lib/components/billing/estimatedTotal.svelte
```diff
@@ -1,144 +0,0 @@
...
-                {#if item.value > 0}
-                    <span class="u-flex u-main-space-between">
-                        <p class="text">{item.label}</p>
-                        <p class="text">{formatCurrency(item.value)}</p>
-                    </span>
-                {/if}
-            {/each}
-            {#each estimation.discounts ?? [] as item}
-                <DiscountsApplied {fixedCoupon} bind:couponData {...item} />
-            {/each}
```
stnguyen90, line 109: This is missing.

PR #1962, opened 2025-06-01
Fix datetime

ItzNotABug:
## What does this PR do?

Screenshot 2025-06-01 at 2 39 50 PM

Screenshot 2025-06-01 at 2 40 08 PM

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/src/lib/components/dualTimeView.svelte
+++ b/src/lib/components/dualTimeView.svelte
@@ -97,7 +97,11 @@
         <Layout.Stack gap="s" alignContent="flex-start">
             <!-- `Raw time` as per design -->
             <Typography.Caption color="--fgcolor-neutral-tertiary" variant="400">
-                {timeToString}
+                {#if $$slots.title}
+                    <slot name="title" />
+                {:else}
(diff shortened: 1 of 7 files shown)


src/lib/components/dualTimeView.svelte
```diff
@@ -97,7 +97,11 @@
         <Layout.Stack gap="s" alignContent="flex-start">
             <!-- `Raw time` as per design -->
             <Typography.Caption color="--fgcolor-neutral-tertiary" variant="400">
-                {timeToString}
+                {#if $$slots.title}
+                    <slot name="title" />
+                {:else}
+                    {timeToString}
+                {/if}
```
Meldiron, line 104: Should this be in else? Should we not show time if it got title?

ItzNotABug: it is in else. if you don't use a slot, we show the default time difference like - `Just now`, `5 mins ago`, etc. You can use custom content when passing a slot.

PR #1965, opened 2025-06-02
Only show sidebar padding on project pages

ernstmul:
## What does this PR do?
Remove the unwanted padding when not on a project related page.

Before (watch the footer padding):
image

After:
image

### DIFF
--- a/src/lib/layout/shell.svelte
+++ b/src/lib/layout/shell.svelte
@@ -113,7 +113,7 @@
     $: state = $isSidebarOpen ? 'open' : 'closed';
 
     let isProjectPage;
-    $: isProjectPage = $page.url.pathname.includes('project-');
+    $: isProjectPage = $page.route.id.includes('project-');
 
     function handleResize() {
         $isSidebarOpen = false;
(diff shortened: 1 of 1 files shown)


src/lib/layout/shell.svelte
```diff
@@ -112,6 +112,9 @@
     let state: undefined | 'open' | 'closed' | 'icons' = 'closed';
     $: state = $isSidebarOpen ? 'open' : 'closed';
 
+    let isProjectPage;
+    $: isProjectPage = $page.url.pathname.includes('project-');
```
ItzNotABug: maybe use `page.route.id?.includes`

ernstmul: Done!

src/lib/layout/shell.svelte
```diff
@@ -112,6 +112,9 @@
     let state: undefined | 'open' | 'closed' | 'icons' = 'closed';
     $: state = $isSidebarOpen ? 'open' : 'closed';
 
+    let isProjectPage;
+    $: isProjectPage = $page.route.id.includes('project-');
```
ItzNotABug:
```suggestion
    $: isProjectPage = $page.route?.id.includes('project-');
```
its nullable type so just in case.

ItzNotABug (changes requested):

PR #1966, opened 2025-06-02
Update: cleanups, improvements

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


src/routes/(console)/+layout.svelte
```diff
@@ -62,11 +66,56 @@
...
+                Query.orderDesc('$updatedAt')
+            ])
+            .then((res) => {
+                latestProjects = res.projects ?? [];
+                for (const project of latestProjects) {
+                    project.region ??= 'default';
+                }
+            })
+            .finally(() => (loadingProjectsForShell = false));
+    }
```
TorstenDittmann: I think we should keep all data loadings in +page/layout.ts files - otherwise they are extrmely hard to invalidate. if you need to do it non-block - you can return a promise from there 👍🏻

ItzNotABug: addressed!

src/routes/(console)/+layout.ts
```diff
@@ -1,71 +1,77 @@
...
         plansInfo,
         roles: [],
         scopes: [],
-        projects: projects,
+        preferences,
+        currentOrgId,
+        organizations,
         currentProjectId: params.project ?? '',
-        organizations: organizations
+        // that this is an unresolved promise
```
ItzNotABug:
```suggestion
        // this is an unresolved promise
```

src/lib/layout/shell.svelte
```diff
@@ -19,16 +19,21 @@
...
+    export let projectsPromise: Promise<Models.ProjectList> = Promise.resolve({
+        total: 0,
+        projects: []
+    });
 
     let yOnMenuOpen: number;
     let showContentTransition = false;
     let timeoutId: ReturnType<typeof setTimeout>;
 
+    $: loadedProjects = [];
```
TorstenDittmann: this is not reactive that way

src/lib/layout/shell.svelte
```diff
@@ -127,6 +132,20 @@
...
+            return {
+                name: project?.name,
+                $id: project.$id,
+                isSelected: project.$id === $page.params.project,
+                region: project.region,
+                platformCount: project.platforms.length,
+                pingCount: project.pingCount
+            };
+        }) satisfies NavbarProject[];
+    });
```
TorstenDittmann: instead of doing this, we should just pass around the promise

src/routes/(console)/+layout.ts
```diff
@@ -1,71 +1,77 @@
...
         roles: [],
         scopes: [],
-        projects: projects,
+        preferences,
+        currentOrgId,
+        organizations,
         currentProjectId: params.project ?? '',
-        organizations: organizations
+        // this is an unresolved promise
+        projectsPromise: loadNonBlocking(currentOrgId)
```
TorstenDittmann: why do we need a function for that? we can just do all of it before

src/routes/(console)/organization-[organization]/+layout.ts
```diff
@@ -10,51 +10,45 @@ import { headerAlert } from '$lib/stores/headerAlert';
...
-                    });
-                }
+                loadFailedInvoices(params.organization);
             }
         }
-        const prefs = await sdk.forConsole.account.getPrefs();
         if (prefs.organization !== params.organization) {
             const newPrefs = { ...prefs, organization: params.organization };
-            sdk.forConsole.account.updatePrefs(newPrefs);
+            void sdk.forConsole.account.updatePrefs(newPrefs);
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/+layout.ts]: i dont think we need the void, we can just call it

src/routes/(console)/organization-[organization]/+layout.ts
```diff
@@ -65,9 +59,22 @@ export const load: LayoutLoad = async ({ params, depends }) => {
             scopes
         };
     } catch (e) {
-        const prefs = await sdk.forConsole.account.getPrefs();
         const newPrefs = { ...prefs, organization: null };
-        sdk.forConsole.account.updatePrefs(newPrefs);
+        void sdk.forConsole.account.updatePrefs(newPrefs);
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/+layout.ts]: again, we dont need a void if not assigning

src/routes/(console)/organization-[organization]/+layout.ts
```diff
@@ -65,9 +59,22 @@ export const load: LayoutLoad = async ({ params, depends }) => {
...
+    failedInvoice.load(teamId).then(() => {
+        if (get(failedInvoice)) {
+            headerAlert.add({
+                show: true,
+                component: ProjectsAtRisk,
+                id: 'projectsAtRisk',
+                importance: 1
+            });
+        }
+    });
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/+layout.ts:83]: this function is a bit weird, we load the invoice and then `get` the store. why dont we just get it in the promise chain?

src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
```diff
@@ -61,27 +66,40 @@
...
+            <Table.Root let:root {columns}>
                 <svelte:fragment slot="header" let:root>
                     <Table.Header.Cell column="dueDate" {root}>Due date</Table.Header.Cell>
                     <Table.Header.Cell column="status" {root}>Status</Table.Header.Cell>
                     <Table.Header.Cell column="amount" {root}>Amount due</Table.Header.Cell>
                     <Table.Header.Cell column="action" {root} />
                 </svelte:fragment>
+
+                {#if isLoadingInvoices}
+                    {#each Array.from({ length: 2 }) as _}
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte]: i would just hardcode it, the Array.from approach requires type workarounds and values are always empty.

src/routes/(console)/project-[region]-[project]/+layout.ts
```diff
@@ -11,16 +11,20 @@ import { get } from 'svelte/store';
...
 import PaymentFailed from '$lib/components/billing/alerts/paymentFailed.svelte';
 import { loadAvailableRegions } from '$routes/(console)/regions';
-import type { Organization } from '$lib/stores/organization';
+import { roles as rolesStore, scopes as scopesStore } from '$lib/stores/roles';
+import { organization as orgStore, type Organization } from '$lib/stores/organization';
 
 export const load: LayoutLoad = async ({ params, depends }) => {
     depends(Dependencies.PROJECT);
     let currentPlan: Plan = null;
+    const orgFromStore: Organization = get(orgStore);
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/+layout.ts]: why do we get org from store? there are race conditions from that are very unreliable. we should only need to fetch it once.

src/routes/(console)/project-[region]-[project]/+layout.ts
```diff
@@ -32,23 +36,22 @@ export const load: LayoutLoad = async ({ params, depends }) => {
             });
         }
         await preferences.loadTeamPrefs(project.teamId);
+
+        // get values from saved stores if available!
+        const [rolesStoreValue, scopesStoreValue] = [get(rolesStore), get(scopesStore)];
+        const areBothAvailable = !!(rolesStoreValue.length && scopesStoreValue.length);
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/+layout.ts]: also will create weird conditions, because potentially they are wrongly invalidated.

in short we should never need to call stores in those files.

src/routes/(console)/project-[region]-[project]/+layout.ts
```diff
@@ -64,3 +67,17 @@ export const load: LayoutLoad = async ({ params, depends }) => {
...
+        if (get(failedInvoice)) {
+            headerAlert.add({
+                show: true,
+                component: PaymentFailed,
+                id: 'paymentFailed',
+                importance: 1
+            });
+        }
+    });
+}
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/+layout.ts:74]: dont we do the same in another file?

src/routes/(console)/project-[region]-[project]/+page.ts
```diff
@@ -4,8 +4,7 @@ import type { PageLoad } from './$types';
 import { hasOnboardingDismissed } from '$lib/helpers/onboarding';
 
 export const load: PageLoad = async ({ params }) => {
-    if (hasOnboardingDismissed(params.project)) {
-        redirect(302, `${base}/project-${params.region}-${params.project}/overview`);
-    }
-    redirect(302, `${base}/project-${params.region}-${params.project}/get-started`);
+    const path = !hasOnboardingDismissed(params.project) ? 'get-started' : 'overview/platforms';
+
+    redirect(302, `${base}/project-${params.region}-${params.project}/${path}`);
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/+page.ts]: thats actually harder to read imo :D

src/routes/(console)/project-[region]-[project]/functions/templates/+page.ts
```diff
@@ -12,40 +12,38 @@ export const load: PageLoad = async ({ url, depends, params }) => {
...
-        .functions.listTemplates(undefined, undefined, 100);
-
-    const [runtimes, useCases] = templates.reduce(
-        ([rt, uc], next) => {
-            next.runtimes.forEach((runtime) => rt.add(runtime.name));
-            next.useCases.forEach((useCase) => uc.add(useCase));
-            return [rt, uc];
-        },
-        [new Set<string>(), new Set<string>()]
+    const localSdkInstance = sdk.forProject(params.region, params.project);
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/functions/templates/+page.ts]: not sure we need such a helper var

src/routes/(console)/project-[region]-[project]/functions/templates/+page.ts
```diff
@@ -54,6 +52,6 @@ export const load: PageLoad = async ({ url, depends, params }) => {
         useCases,
         search,
         templates,
-        functions: await sdk.forProject(params.region, params.project).functions.list()
+        functions: await localSdkInstance.functions.list()
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/functions/templates/+page.ts]: how come we need functions here anyway? we are in templates

src/routes/(console)/project-[region]-[project]/overview/+page.ts
```diff
@@ -5,9 +5,8 @@ import type { PageLoad } from './$types';
...
 export const load: PageLoad = async ({ params }) => {
-    if (get(selectedTab) === 'keys') {
-        redirect(302, `${base}/project-${params.region}-${params.project}/overview/keys`);
-    } else {
-        redirect(302, `${base}/project-${params.region}-${params.project}/overview/platforms`);
-    }
+    const tab = get(selectedTab);
+    const subroute = ['keys', 'dev-keys'].includes(tab) ? tab : 'platforms';
+
+    redirect(302, `${base}/project-${params.region}-${params.project}/overview/${subroute}`);
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/overview/+page.ts]: i think a switch is way more readable here.

sometimes repetition is not that bad - because it makes it extremely easy to understand whats going on.

src/routes/(console)/organization-[organization]/+layout.ts
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/+layout.ts:83]: no idea. its always been like that.

src/routes/(console)/project-[region]-[project]/+layout.ts
Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/+layout.ts]: to not fetch from api if it already exists.

src/routes/(console)/project-[region]-[project]/+layout.ts
Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/+layout.ts:74]: yes, but a different component for alert.

src/routes/(console)/project-[region]-[project]/functions/templates/+page.ts
Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/functions/templates/+page.ts]: don't think we need em. better to remove.

src/routes/(console)/organization-[organization]/+layout.ts
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/+layout.ts:83]: not removing this atm, the function has references across multiple components. I'll relay this to billing in the meantime. this feels a bit out of scope for this one.

src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte]: feels a bit cleaner and why do we care about type, or the value? its X amount of rows loading skeletons.

ItzNotABug:
TODO: load region based console variable early rather than per project.

src/lib/components/breadcrumbs.svelte
```diff
@@ -14,20 +22,15 @@
...
-        isSelected: boolean;
-        region: string;
-    };
     type Organization = {
         name: string;
         $id: string;
         tierName: string;
         isSelected: boolean;
-        projects: Array<Project>;
+        projects: Promise<Models.ProjectList>;
```
TorstenDittmann: how come we need this? organization shouldnt need a connection to projects for that ui

src/lib/components/breadcrumbs.svelte
```diff
@@ -14,20 +22,15 @@
     import { base } from '$app/paths';
     import { newOrgModal } from '$lib/stores/organization';
     import { Click, trackEvent } from '$lib/actions/analytics';
-    import { page } from '$app/stores';
+    import type { Models } from '@appwrite.io/console';
+    import { project } from '$routes/(console)/project-[region]-[project]/store';
```
TorstenDittmann: lets try avoid importing route states in the components, rather just have a property for this 👍🏻

src/lib/components/breadcrumbs.svelte
```diff
@@ -77,6 +78,11 @@
         } else newOrgModal.set(true);
     }
 
+    async function loadProjectsFromPromise() {
+        loadedProjects = await selectedOrg.projects;
+        return loadedProjects;
+    }
```
TorstenDittmann: mhh, I think we should get rid of the big blocks on the bottom.

lets use a pattern like `async function createBottomSheet(...)` and use it like `$: projectsBottomSheet = createBottomSheet(...)`.

that wait we can just `{#await projectsBottomSheet}`

src/routes/(console)/+layout.ts
```diff
@@ -1,71 +1,69 @@
...
     };
 };
+
+async function loadPromises() {
+    return Promise.all([
+        sdk.forConsole.account.getPrefs(),
+        isCloud ? sdk.forConsole.billing.getPlansInfo() : Promise.resolve(null),
+        isCloud ? sdk.forConsole.billing.listOrganization() : sdk.forConsole.teams.list()
+    ]);
+}
```
TorstenDittmann: not sure its worth putting it in a function when its only ever used once 👍🏻

src/routes/(console)/+layout.ts
```diff
@@ -1,71 +1,69 @@
...
-            Query.orderDesc('$updatedAt')
-        ]);
-        projects = orgProjects.projects.length > 0 ? orgProjects.projects : [];
-    }
-
-    // set `default` if no region!
-    for (const project of projects) {
-        project.region ??= 'default';
-    }
+    sdk.forConsole.console.variables().then((vars) => consoleVariables.set(vars));
```
TorstenDittmann: using stores in +layout doesnt work that well 👍🏻

src/routes/(console)/+layout.ts
```diff
@@ -1,71 +1,69 @@
...
+        (organizations.teams.length > 0 ? organizations.teams[0].$id : undefined);
 
-    const organizations = !isCloud
-        ? await sdk.forConsole.teams.list()
-        : await sdk.forConsole.billing.listOrganization();
+    fetch(`${endpoint}/health/version`, {
+        headers: { 'X-Appwrite-Project': project }
+    })
+        .then((res) => res.json().catch(() => null))
+        .then((data) => version.set(data?.version));
```
TorstenDittmann: same here we stores

TorstenDittmann: if we wanna put async data into stores, it should happen with the store, in the load functions 👍🏻

src/routes/(console)/organization-[organization]/billing/+page.ts
```diff
@@ -49,12 +51,19 @@ export const load: PageLoad = async ({ parent, depends }) => {
         // ignore error
     }
 
+    const areCreditsSupported = isCloud
+        ? (get(currentPlan)?.supportsCredits ??
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/billing/+page.ts]: cant we get the current plan from the `parent()`?

src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
Comment (TorstenDittmann) [reply src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte]: because svelte 5 actually wants us to have keys for each. also a `_` is not that clean :P

we would need to do:

```suggestion
                    {#each Array.from({ length: 2 }).keys() as index (index)}
```

to have it clean ;-)

src/lib/components/breadcrumbs.svelte
```diff
@@ -96,92 +99,89 @@
...
     function onResize() {
         if ((organisationBottomSheetOpen || projectsBottomSheetOpen) && !$isSmallViewport) {
             organisationBottomSheetOpen = false;
             projectsBottomSheetOpen = false;
         }
     }
+
+    $: projectsBottomSheet = createProjectsBottomSheet();
+
+    $: organizationsBottomSheet = createOrganizationBottomSheet();
```
TorstenDittmann: this is not reactive. we need to pass the props we want it to be reactive with. otherwise its not difference to a `const`

src/lib/layout/shell.svelte
```diff
@@ -19,11 +18,14 @@
...
-    export let showSideNavigation = false;
     export let showHeader = true;
     export let showFooter = true;
-    export let loadedProjects: Array<NavbarProject> = [];
+    export let showSideNavigation = false;
     export let selectedProject: Models.Project | null = null;
+    export let projectsPromise: Promise<Models.ProjectList> = Promise.resolve({
+        total: 0,
+        projects: []
+    });
```
TorstenDittmann: we probably should be a bit more explicit, it should be a promise or undefined. otherwise we dont really know if it was an actual promise or the mock here

src/routes/(console)/+layout.svelte
```diff
@@ -323,6 +314,19 @@
...
+            Query.equal(
+                'teamId',
+                data.currentOrgId ?? currentOrganizationId ?? page.params.organization
+            ),
+            Query.limit(5),
+            Query.orderDesc('$updatedAt')
+        ]);
+    }
+
+    $: projectsPromise = getProjectsPromise();
```
TorstenDittmann: this is not react, since getProjectsPromise never changes

TorstenDittmann: you coud do this for example:

```js
        $:projects = sdk.forConsole.projects.list([
            Query.equal(
                'teamId',
                data.currentOrgId ?? currentOrganizationId ?? page.params.organization
            ),
            Query.limit(5),
            Query.orderDesc('$updatedAt')
        ]);
```

src/routes/(console)/+layout.ts
```diff
@@ -1,71 +1,55 @@
...
-    }
-    if (currentOrgId) {
-        const orgProjects = await sdk.forConsole.projects.list([
-            Query.equal('teamId', currentOrgId),
-            Query.limit(5),
-            Query.orderDesc('$updatedAt')
+    const [preferences, plansArray, organizations, versionData, consoleVariables] =
+        await Promise.all([
+            sdk.forConsole.account.getPrefs(),
+            isCloud ? sdk.forConsole.billing.getPlansInfo() : Promise.resolve(null),
```
TorstenDittmann:
```suggestion
            isCloud ? sdk.forConsole.billing.getPlansInfo() : null,
```

this should work too, no difference 👍🏻

src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte
```diff
@@ -21,32 +21,39 @@
...
 
     const tier = data?.plan
         ? (data.plan.$id as Tier)
         : (data?.currentInvoice?.plan ?? $organization?.billingPlan);
+
     const plan = data?.plan ?? undefined;
 
     $: projects = (data.organizationUsage as OrganizationUsage).projects;
 
+    let usageProjects: { [key: string]: UsageProjectInfo } = {};
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte]:
```suggestion
    let usageProjects: Record<string, UsageProjectInfo> = {};
```
can be a cleaner approach for the type 👍🏻

src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte
```diff
@@ -21,32 +21,39 @@
...
             name: 'Writes',
             value: data.organizationUsage.databasesWrites.reduce(
-                (sum, singleDay) => sum + singleDay.value,
+                (sum: number, singleDay: { date: string; value: number }) => sum + singleDay.value,
                 0
             )
         }
     ];
+
+    onMount(async () => (usageProjects = await data.projectsPromise));
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte]: why dont we pass the promise and await where we actually need it? 👀

src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.ts
```diff
@@ -49,43 +50,57 @@ export const load: PageLoad = async ({ params, parent }) => {
...
-
     return {
         plan,
         invoices,
-        projects,
         currentInvoice,
         organizationMembers,
-        organizationUsage: usage
+        organizationUsage: usage,
+        projectsPromise: getUsageProjectsPromise(usage)
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.ts]: we should just call it projects, we dont have to append promise everytime 👍🏻

src/routes/(console)/project-[region]-[project]/overview/+page.ts
```diff
@@ -5,9 +5,18 @@ import type { PageLoad } from './$types';
...
+        case 'keys':
+        case 'dev-keys':
+            subroute = tab;
+            break;
+        default:
+            subroute = 'platforms';
+            break;
     }
+
+    redirect(302, `${base}/project-${params.region}-${params.project}/overview/${subroute}`);
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/overview/+page.ts]: why not just this? we can save a lot of variables.

```suggestion
    const tab = get(selectedTab);

    switch (tab) {
        case 'keys':
        case 'dev-keys':
            redirect(302, `${base}/project-${params.region}-${params.project}/overview/${tab}`);
        default:
            redirect(302, `${base}/project-${params.region}-${params.project}/overview/platforms`);
    }
```

maybe also we shouldnt rely on the selectedTab variable to actually be the right path.

src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte]: Its used on multiple `ProjectBreakdown` components based on the type of service. So it makes sense to load all of them here.

src/routes/(console)/project-[region]-[project]/overview/+page.ts
Comment (ItzNotABug) [reply src/routes/(console)/project-[region]-[project]/overview/+page.ts]: the duplicacy of redirect multiple times was just bugging me so I did the above thing. this is OK as well.

src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/usage/[[invoice]]/+page.svelte]: Also it adds a bit of redundant logic to maintain a promise and a resolved var with value under `ProjectBreakdown`. Doesn't feel like the component's responsibility.

src/routes/(console)/project-[region]-[project]/overview/header.svelte
```diff
@@ -2,26 +2,35 @@
...
     import { user } from '$lib/stores/user';
     import { isSmallViewport } from '$lib/stores/viewport';
-    import { Dependencies } from '$lib/constants';
     import { trackEvent } from '$lib/actions/analytics';
+
+    function dismissOnboarding() {
+        setHasOnboardingDismissed($project.$id);
+        trackEvent('onboarding_hub_platform_dismiss');
+        // TODO: @ernst, do we need this invalidate?
+        // await invalidate(Dependencies.ORGANIZATION);
```
Comment (ernstmul) [inline src/routes/(console)/project-[region]-[project]/overview/header.svelte]: I checked, it can be removed. It was so the "dismiss onboarding" would disappear from the side bar. But it now also works without this step due to changes made later on. So safe to remove this.

src/routes/(console)/+layout.ts
```diff
@@ -1,71 +1,66 @@
...
+    const [preferences, plansArray, orgs, versionData, consoleVariables] = await Promise.all([
+        sdk.forConsole.account.getPrefs(),
+        isCloud ? sdk.forConsole.billing.getPlansInfo() : null,
+        // reuse from parent if available, else load fresh.
+        // know that the parent loads these orgs conditionally!
+        !organizations
+            ? isCloud
+                ? sdk.forConsole.billing.listOrganization()
+                : sdk.forConsole.teams.list()
+            : organizations,
```
TorstenDittmann: do we even need this if parent has it already? the dependency is invalidated anyway when its needed 👍🏻

TorstenDittmann (approved):

src/routes/(console)/+layout.ts
ItzNotABug: done.

ItzNotABug:
**Approved by the lead wizard, but DO NOT merge yet.**

Yes, it's green. No, you still can't touch it. You'll thank me later.

ItzNotABug:
**Note to self: No more commits here. Log other issues on Linear and handle them separately.**

src/lib/components/navbar.svelte
```diff
@@ -134,13 +138,14 @@
             class="only-desktop">
             <img src={logo.src} alt={logo.alt} />
         </a>
-        <Breadcrumbs {organizations} />
-        {#if selectedProject && selectedProject.pingCount === 0}
+        <Breadcrumbs {organizations} {projects} {currentProject} />
+        {#if page.route?.id?.includes('project-') && currentProject && currentProject.pingCount === 0}
```
TorstenDittmann:
```suggestion
        {#if page.route?.id?.includes('/project-[region]-[project]') && currentProject && currentProject.pingCount === 0}
```

lets be a bit more explicit, otherwise any `project-` in the url could trigger it 👍🏻

src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
```diff
@@ -63,27 +67,42 @@
...
     });
+
+    const columns = $derived.by(() => {
+        return [
+            { id: 'dueDate', width: { min: 120 } },
+            { id: 'status', width: { min: hasPaymentError ? 200 : 100 } },
+            { id: 'amount', width: { min: 120 } },
+            { id: 'action', width: 40 }
+        ];
+    });
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte]:
```suggestion
    const columns = $derived([
        { id: 'dueDate', width: { min: 120 } },
        { id: 'status', width: { min: hasPaymentError ? 200 : 100 } }
        { id: 'amount', width: { min: 120 } },
        { id: 'action', width: 40 }
    ]);
```
this works too right?

src/routes/(console)/organization-[organization]/change-plan/+page.svelte
```diff
@@ -35,15 +36,17 @@
 
     export let data;
 
+    let paymentMethods: PaymentList = null;
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/change-plan/+page.svelte]: do we still need this var then?

src/routes/(console)/organization-[organization]/change-plan/+page.svelte
```diff
@@ -35,15 +36,17 @@
...
     let selectedPlan: BillingPlan = data.plan as BillingPlan;
-    let selectedCoupon: Partial<Coupon> | null = data.coupon;
     let previousPage: string = base;
     let showExitModal = false;
     let formComponent: Form;
     let isSubmitting = writable(false);
     let paymentMethodId: string =
         data.organization.paymentMethodId ??
-        data.paymentMethods.paymentMethods.find((method) => !!method?.last4)?.$id;
+        paymentMethods?.paymentMethods?.find((method) => !!method?.last4)?.$id;
```
Comment (TorstenDittmann) [inline src/routes/(console)/organization-[organization]/change-plan/+page.svelte]: this might not work, since its only populated before payment methods has been fulfilled.

src/routes/(console)/project-[region]-[project]/overview/+page.ts
```diff
@@ -1,13 +1,7 @@
...
 import type { PageLoad } from './$types';
-import { selectedTab } from './store';
 
 export const load: PageLoad = async ({ params }) => {
-    if (get(selectedTab) === 'keys') {
-        redirect(302, `${base}/project-${params.region}-${params.project}/overview/keys`);
-    } else {
-        redirect(302, `${base}/project-${params.region}-${params.project}/overview/platforms`);
-    }
+    return redirect(302, `${base}/project-${params.region}-${params.project}/overview/platforms`);
```
Comment (TorstenDittmann) [inline src/routes/(console)/project-[region]-[project]/overview/+page.ts]:
```suggestion
    redirect(302, `${base}/project-${params.region}-${params.project}/overview/platforms`);
```
redirect returns `never`

src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/billing/paymentHistory.svelte:78]: yepp

src/routes/(console)/organization-[organization]/change-plan/+page.svelte
Comment (ItzNotABug) [reply src/routes/(console)/organization-[organization]/change-plan/+page.svelte]: We kinda do for cases if the `data.organization.paymentMethodId` is not available.

PR #1882, opened 2025-05-23
string encryption in create attribute

ArnabChatterjee20k:
## What does this PR do?
[Screencast from 2025-05-23 19-47-58.webm](https://github.com/user-attachments/assets/19ed41f8-a06a-4a08-bfd2-3e761449b869)

(Provide a description of what this PR does.)

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work.)

## Related PRs and Issues

(If this PR is related to any other PR or resolves any issue or related to any issue link all related PR and issues here.)

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

(Write your answer here.)

PR #2049, opened 2025-06-26
fix: improve modal focus

HarshMN2345:
## What does this PR do?
Implements a dual focus strategy for modals using both `autofocus` attribute and `onMount` focus handling. This ensures reliable input focus across different scenarios (initial load and dynamic content) and improves keyboard accessibility.

## Test Plan
1. Initial page load:
   - Open any modal with input field
   - Verify first input is focused automatically

2. Dynamic content:
   - Open modals through user interactions
   - Verify input focus works when modal opens
   - Test with keyboard navigation

3. Browser testing:
   - Tested on Chrome, Firefox, Safari
   - Verified no double-focus effects occur

## Related PRs and Issues
Fixes #1882 - Modal focus not working consistently

Yes

### DIFF
--- a/src/lib/components/feedback/evaluation.svelte
+++ b/src/lib/components/feedback/evaluation.svelte
@@ -1,17 +1,7 @@
 <script lang="ts">
-    import { onMount } from 'svelte';
     import { Tag } from '@appwrite.io/pink-svelte';
 
     export let value: number = null;
-
-    onMount(() => {
-        setTimeout(() => {
-            const firstButton = document.querySelector('[data-rating="0"]') as HTMLElement;
-            if (firstButton) {
-                firstButton.focus();
-            }
-        }, 150);
-    });
 </script>
 
 <fieldset class="u-margin-block-start-8 u-min-width-0">
--- a/src/lib/components/feedback/evaluation.svelte
+++ b/src/lib/components/feedback/evaluation.svelte
@@ -1,7 +1,17 @@
 <script lang="ts">
+    import { onMount } from 'svelte';
     import { Tag } from '@appwrite.io/pink-svelte';
 
     export let value: number = null;
+
+    onMount(() => {
+        setTimeout(() => {
+            const firstButton = document.querySelector('[data-rating="0"]') as HTMLElement;
+            if (firstButton) {
+                firstButton.focus();
+            }
+        }, 150);
+    });
 </script>
 
 <fieldset class="u-margin-block-start-8 u-min-width-0">
@@ -10,7 +20,12 @@
         style="padding-block: 0.13rem">
         {#each Array(11) as _, i}
             <li>
-                <Tag size="m" selected={value === i} on:click={() => (value = i)}>
+                <Tag
+                    size="m"
+                    selected={value === i}
+                    autofocus={i === 0}
+                    data-rating={i}
+                    on:click={() => (value = i)}>
                     {i}
                 </Tag>
             </li>
--- a/src/lib/components/feedback/feedbackGeneral.svelte
+++ b/src/lib/components/feedback/feedbackGeneral.svelte
@@ -1,11 +1,22 @@
 <script lang="ts">
+    import { onMount } from 'svelte';
     import { InputTextarea } from '$lib/elements/forms';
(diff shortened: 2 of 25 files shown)


src/lib/components/feedback/evaluation.svelte
ItzNotABug, line 1: Lets try to avoid `onMount()` hooks along with `document.querySelector`. Its gets very messy to maintain and isn't a good DX as well. See if the `autofocus` on the `Input` components itself work. If they don't we might need to raise a PR on Pink2 side.

HarshMN2345:
Hey @ItzNotABug,<br>Noticed that `autofocus` isn’t working as expected in components like Feedback, Evaluate, and Create Organization — turns out Pink2’s `Input` doesn’t forward the prop properly. We could rely on `onMount` for now as a workaround in places where initial focus is important, though it's a bit messy. Ideally, Pink2 should support native `autofocus`, and once that's fixed, we can drop the manual focus handling.

src/routes/(console)/project-[region]-[project]/auth/createUser.svelte
```diff
@@ -56,7 +56,7 @@
 </script>
 
 <Modal title="Create user" {error} bind:show={showCreate} onSubmit={create}>
-    <InputText id="name" label="Name" placeholder="Enter name" autofocus={true} bind:value={name} />
+    <InputText autofocus={true} id="name" label="Name" placeholder="Enter name" bind:value={name} />
```
Comment (ItzNotABug) [inline src/routes/(console)/project-[region]-[project]/auth/createUser.svelte]: can be just `autofocus`.

src/routes/(console)/project-[region]-[project]/messaging/create-[type]/(type)/email.svelte
```diff
@@ -106,6 +106,7 @@
                         id="subject"
                         label="Subject"
                         required
+                        autofocus={true}
```
Comment (ItzNotABug) [inline src/routes/(console)/project-[region]-[project]/messaging/create-[type]/(type)/email.svelte:109]: for truthy values, just use the prop name, it works ok. goes for all the others as well.

src/lib/components/feedback/feedbackGeneral.svelte
```diff
@@ -1,11 +1,22 @@
...
     import { feedbackData } from '$lib/stores/feedback';
+
+    onMount(() => {
+        setTimeout(() => {
+            const textarea = document.getElementById('feedback') as HTMLTextAreaElement;
+            if (textarea) {
+                textarea.focus();
+            }
+        }, 150);
+    });
```
ItzNotABug: we should still avoid this.
the reason feedback doesn't get focus is because there are 2 inputs shown, one for mobile, one for desktop. the mobile one is hidden but still grabs the focus. check/search for the `MobileFeedbackModal` component. We also have a utility store called - `isSmallViewport`, you can use that to figure out the viewport and conditionally render the components so the focus is only on one of the input texts 👍.

src/lib/components/sidebar.svelte
```diff
@@ -269,7 +269,9 @@
                                 >Feedback
                             </Button.Button>
                             <svelte:fragment slot="other">
-                                <MobileFeedbackModal />
+                                {#if $isSmallViewport}
```
ItzNotABug: both these feedback and support sections are for mobile only right? instead of only using the small view port check on feedback, lets move the check on an upper level.

src/lib/components/sidebar.svelte
```diff
@@ -329,7 +331,9 @@
                                     ><span>Feedback</span>
                                 </Button.Button>
                                 <svelte:fragment slot="other">
-                                    <MobileFeedbackModal />
+                                    {#if $isSmallViewport}
+                                        <MobileFeedbackModal />
+                                    {/if}
```
ItzNotABug: same as above if the mobile only section applies here.

src/lib/components/sidebar.svelte
```diff
@@ -255,7 +255,8 @@
                         </Tooltip>
                     </div>
                 </Layout.Stack>
-            {:else}
+            {/if}
+            {#if $isSmallViewport}
```
ItzNotABug: I think this was shown on an `else` block. should be `{:else if}` and work correctly as per the previous logic.

PR #2060, opened 2025-06-29
Fix: payment methods not updated after adding it

ItzNotABug:
## What does this PR do?

Adding a new card required a page reload on `change-plan` route due to the payment methods being loaded in svelte context which didn't trigger a reload without a reactive update. This PR fixes that and directly updates the underlying payment list modal to avoid a skeleton loader flash if we used a streamed promise or reloaded via svelte context await.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/e2e/journeys/upgrade-free-tier.spec.ts
+++ b/e2e/journeys/upgrade-free-tier.spec.ts
@@ -11,7 +11,12 @@ test('upgrade - free tier', async ({ page }) => {
         await page.waitForURL(/\/organization-[^/]+\/change-plan/);
         await page.locator('input[value="tier-1"]').click();
         await page.getByRole('button', { name: 'add' }).first().click();
+
         await enterCreditCard(page);
+
+        // wait for a second after adding a card to update the UI.
+        await page.waitForSelector('button#method[role="combobox"]');
+
         // skip members
         await page.getByRole('button', { name: 'change plan' }).click();
         await page.waitForURL(/\/console\/project-(?:[a-z0-9]+-)?([^/]+)\/get-started/);


src/lib/components/billing/selectPaymentMethod.svelte
```diff
@@ -20,8 +20,19 @@
...
-        invalidate(Dependencies.UPGRADE_PLAN);
-        invalidate(Dependencies.CREATE_ORGANIZATION);
+
+        if (value) {
+            methods = {
+                ...methods,
+                total: methods.total + 1,
+                paymentMethods: [...methods.paymentMethods, event.detail]
+            };
+        }
```
ItzNotABug, line 30: directly adds the card to the passed on object. this is done to avoid a flash of layout shift if we used `invalidate(Dependencies.PAYMENT_METHODS)` because we use a Skeleton loader, the load is pretty quick so it would look very off. Doing this way, there's just a UI swap instead of a network request.

src/routes/(console)/+layout.svelte
```diff
@@ -316,15 +316,19 @@
...
-            'teamId',
-            // id from page params ?? id from store ?? id from preferences
-            page.params.organization ?? currentOrganizationId ?? data.currentOrgId
-        ),
-        Query.limit(5),
-        Query.orderDesc('$updatedAt')
-    ]);
+    $: isOnOnboarding = page.route?.id?.includes('/(console)/onboarding');
+
+    $: projects = isOnOnboarding
```
ItzNotABug, line 321: avoids a failed network request when there are no orgs and the user is on onboarding path.

src/routes/(console)/onboarding/create-organization/+page.svelte
```diff
@@ -47,6 +47,10 @@
             if (organization) {
                 loadAvailableRegions(organization?.$id).then();
                 await goto(`${base}/organization-${organization.$id}`);
+
+                // fixes an edge case where
+                // the org is not available for some reason!
+                await invalidate(Dependencies.CREATE_ORGANIZATION);
```
ItzNotABug, line 53: this is because we use `const { organizations } = await parent()`, which is loaded at the top level `+layout.ts`. Children below it don't load it so the first project creation can fail sometimes for billing checks because there's no org in the list. This just invalidates and loads all of them again. Perf cost is very minimal.

src/routes/(console)/organization-[organization]/change-plan/+page.svelte
```diff
@@ -338,12 +338,12 @@
...
                     </Fieldset>
-                {:then paymentMethods}
+                {:then}
                     <Fieldset legend="Payment">
                         <SelectPaymentMethod
-                            methods={paymentMethods}
+                            bind:taxId
                             bind:value={paymentMethodId}
-                            bind:taxId>
+                            bind:methods={paymentMethods}>
```
Comment (ItzNotABug) [inline src/routes/(console)/organization-[organization]/change-plan/+page.svelte:346]: binds in order to update/mutate the object.

lohanidamodar (approved):
Looks good to me, make sure it's tested on staging

PR #2085, opened 2025-07-06
Fix: Prevent Modal Dropdown Truncation on Mobile

HarshMN2345:
### Problem
On mobile devices, dropdowns (such as the Columns dropdown in the Adjustments modal) were being truncated or clipped due to the modal’s content area having restricted overflow.

### Solution
CSS-only fix:

- Added a global CSS rule in `src/lib/components/confirm.svelte`
- Allows dropdowns and overlays inside modals to overflow the modal content area
- Ensures dropdowns are fully visible and scrollable on mobile devices

### Reverted Changes
- All previous JavaScript or logic-based changes related to dropdown max-height or mobile-specific handling have been reverted
- This PR contains only the CSS fix mentioned above

### Result
- Dropdowns inside modals are now fully visible and usable across all devices, including mobile
- No other logic or structural code was changed
- This is a minimal and focused fix addressing the truncation issue
### Before and After
Screenshot 2025-07-06 144534
Screenshot 2025-07-06 145447

### DIFF
--- a/src/lib/components/confirm.svelte
+++ b/src/lib/components/confirm.svelte
@@ -68,3 +68,9 @@
         </svelte:fragment>
     </Dialog>
 </Form>
+
+<style>
+    :global(dialog section) {
+        overflow: visible !important;
+    }
+</style>
--- a/src/lib/components/confirm.svelte
+++ b/src/lib/components/confirm.svelte
@@ -68,9 +68,3 @@
         </svelte:fragment>
     </Dialog>
 </Form>
-
-<style>
-    :global(dialog section) {
-        overflow: visible !important;
-    }
-</style>
--- a/src/lib/components/viewSelector.svelte
+++ b/src/lib/components/viewSelector.svelte
@@ -4,6 +4,7 @@
     import type { Column } from '$lib/helpers/types';
     import { Icon, Button } from '@appwrite.io/pink-svelte';
     import { IconViewBoards } from '@appwrite.io/pink-icons-svelte';
+    import { isSmallViewport } from '$lib/stores/viewport';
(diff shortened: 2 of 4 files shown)


src/lib/components/confirm.svelte
ItzNotABug, line 1: the component `ColumnSelector` uses `Popover` which supports a `portal` option, that should be a better solution. A prop to portal should be good for mobile modals because we don't need portal by default.

src/lib/layout/displaySettingsModal.svelte
```diff
@@ -35,7 +36,7 @@
         {#if !hideColumns && $columns?.length}
             <Layout.Stack gap="xs">
                 <Typography.Text>Columns</Typography.Text>
-                <ColumnSelector {columns} {isCustomCollection}>
+                <ColumnSelector {columns} {isCustomCollection} portal={$isSmallViewport}>
```
ItzNotABug, line 39: can just pass `true` here, already inside a modal which is rendered due to `responsiveContainerHeader` > `$isSmallViewport` > `#snippet settingsButton`.

src/lib/components/viewSelector.svelte
```diff
@@ -16,7 +17,7 @@
 </script>
 
 {#if !hideColumns && view === View.Table}
-    <ColumnSelector {columns} {isCustomCollection} {allowNoColumns}>
+    <ColumnSelector {columns} {isCustomCollection} {allowNoColumns} portal={$isSmallViewport}>
```
ItzNotABug, line 20: this might not be needed unless the `ViewSelector` itself is not in a modal or dialog some other upper layered component. Please check and verify.

PR #2086, opened 2025-07-07
fix: Minimalize requests to listRepositories with debounce

HarshMN2345:
## What does this PR do?

This PR optimizes the repositories list component in the Git integration flow.

- Debounces repository search input (300ms) to reduce unnecessary API calls.
- Handles race conditions to ensure only the latest search result is displayed.
- Cleans up unused variables and removes unnecessary try/catch blocks.
- Resolves linter errors and improves overall code readability and efficiency.

---

## Test Plan

- Verified that repository search triggers API requests only after a 300ms pause in typing.
- Confirmed repository list updates correctly when switching installations or clearing the search.
- Ensured only the most recent search result is shown, even if previous requests resolve later.
- Ran `pnpm lint` — no errors or warnings.
- Manually tested UI to confirm no regressions in repo selection/connection.

---

## Related PRs and Issues

N/A

---

## Have you read the Contributing Guidelines on issues?

✅ Yes

### DIFF
--- a/src/lib/components/git/repositories.svelte
+++ b/src/lib/components/git/repositories.svelte
@@ -72,7 +72,7 @@
     });
 
     onDestroy(() => {
-        debouncedLoadRepositories.cancel?.();
+        debouncedLoadRepositories.cancel();
     });
 
     async function loadInstallations() {
@@ -175,7 +175,7 @@
                                 installationsMap.find((entry) => entry.$id === selectedInstallation)
                             );
 
-                            debouncedLoadRepositories.cancel?.();
+                            debouncedLoadRepositories.cancel();
                         }}
                         bind:value={selectedInstallation} />
                     <InputSearch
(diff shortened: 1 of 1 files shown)


src/lib/components/git/repositories.svelte
```diff
@@ -52,6 +53,28 @@
...
+    );
+
+    $effect(() => {
+        if (selectedInstallation && search !== undefined) {
+            debouncedLoadRepositories(selectedInstallation, search);
+        }
+    });
+
+    onDestroy(() => {
+        debouncedLoadRepositories.cancel?.();
```
ItzNotABug: why is `null` check required?

src/lib/components/git/repositories.svelte
```diff
@@ -52,6 +53,28 @@
...
+        async (installationId: string, searchTerm: string) => {
+            isLoadingRepositories = true;
+            try {
+                await loadRepositories(installationId, searchTerm);
+            } finally {
+                isLoadingRepositories = false;
+            }
+        },
+        300
+    );
```
ItzNotABug, line 66: This won't work because the actual request is continuously fired from the `loadRepositories` in the svelte context. This will work but the requests from svelte context below will keep firing on every keystroke due to a change in `search` variable.

Issue #2045, opened 2025-06-25
🐛 Bug Report: Unable to create APNS provider


stnguyen90:
### 👟 Reproduction steps

1. Create a provider
2. Select APNS

### 👍 Expected behavior

It has a file upload for the p8 certificate

### 👎 Actual Behavior

It asks for service JSON:

![Image](https://github.com/user-attachments/assets/3c364c8a-78ec-4fd7-928d-768f11fd480a)

### 🎲 Appwrite version

Appwrite Cloud

### 💻 Operating system

Linux

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

stnguyen90:
It looks like https://github.com/appwrite/console/commit/ec9058ce5c09f60d6606ec66be34a358dd065520 changed the input file and hardcoded it to be a JSON file. We'll look into getting this fixed soon.

scottgrobinson:
Sorry to be "that" person - But... +1

Also unable to upload an FCM P12 JSON file as it throws "Unexpected end of JSON input". Can we raise the priorty of this a little as it means we've not able to use the push notifications on a service we're paying for.

TheArtOfDoing:
Hi, this is marked as closed, but I have a self-hosted Appwrite V1.7.4 running and it still shows json when I try to register the APNS Provider.

PR #2092, opened 2025-07-07
fix: apns provider key file upload input

stnguyen90:
## What does this PR do?

Don't hardcode the upload dropzone to be for service account JSON files only.

Fixes https://github.com/appwrite/console/issues/2045

## Test Plan

Tested locally:

FCM:

image

APNS:

image

## Related PRs and Issues

* https://github.com/appwrite/console/issues/2045

Yes

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte
+++ b/src/routes/(console)/project-[region]-[project]/messaging/providers/settingsFormInput.svelte
@@ -24,8 +24,7 @@
         APNSProviderParams
     } from './store';
     import TooltipPopover from './tooltipPopover.svelte';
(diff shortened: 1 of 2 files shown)


stnguyen90:
One issue I'm noticing is the popover isn't updating correct. For example, after opening the APNS Team ID popover and then switching to FCM, the 1st popover shows the Team ID popover:

image

Any ideas?

ItzNotABug:
> One issue I'm noticing is the popover isn't updating correct. For example, after opening the APNS Team ID popover and then switching to FCM, the 1st popover shows the Team ID popover:
> 
> image
> Any ideas?

could be a reactivity issue. maybe try with a `#key` block with a var that's reactive.

ItzNotABug (approved):

PR #1894, opened 2025-05-26
Remove Promos, Add Sites waitlist

ItzNotABug:
## What does this PR do?

(Provide a description of what this PR does.)

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work.)

## Related PRs and Issues

(If this PR is related to any other PR or resolves any issue or related to any issue link all related PR and issues here.)

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

(Write your answer here.)

PR #2136, opened 2025-07-21
fix: Hide redeploy button when `sourceSize` is 0

hmacr:
## What does this PR do?

Following the pattern established in [DeploymentActionMenu](https://github.com/appwrite/console/blob/main/src/routes/(console)/project-%5Bregion%5D-%5Bproject%5D/sites/(components)/deploymentActionMenu.svelte#L58-L68), this PR hides the `Redeploy` action/button in other places when `sourceSize` is 0.

## Test Plan

Manual QA: verified `Redeploy` button is hidden when deployment `sourceSize` is 0.

## Related PRs and Issues

#1894 where `Redeploy` item was first hidden in `DeploymentActionMenu`.

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/functions/function-[function]/+page.svelte
+++ b/src/routes/(console)/project-[region]-[project]/functions/function-[function]/+page.svelte
@@ -98,7 +98,7 @@
                                 </Button>
                                 <svelte:fragment slot="menu" let:toggle>
                                     <ActionMenu.Root>
-                                        {#if $canWriteFunctions}
+                                        {#if $canWriteFunctions && activeDeployment.sourceSize !== 0}
                                             <ActionMenu.Item.Button
                                                 trailingIcon={IconRefresh}
                                                 on:click={() => {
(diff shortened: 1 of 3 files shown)


src/routes/(console)/project-[region]-[project]/functions/function-[function]/+page.svelte
```diff
@@ -98,7 +98,7 @@
                                 </Button>
                                 <svelte:fragment slot="menu" let:toggle>
                                     <ActionMenu.Root>
-                                        {#if $canWriteFunctions}
+                                        {#if $canWriteFunctions && activeDeployment.sourceSize !== 0}
```
Comment (Meldiron) [inline src/routes/(console)/project-[region]-[project]/functions/function-[function]/+page.svelte]: Hiding it all-together is considered bad UX, as it doesn't explain why it's sometimes disappearing. Also can cause trouble to feature discoverability.

Instead, let's just set it disabled: https://github.com/appwrite/console/blob/8381af6982a1eff132b1e40cccf6bc2ba3d508d4/src/routes/(console)/project-[region]-[project]/functions/function-[function]/(components)/downloadActionMenuItem.svelte#L26-L33

Also, let's add tooltip to explain why it's disabled, It could say something like `Source is empty`.

Please also add tooltip to download button I linked above, it seems to be missing it too.

Comment (hmacr) [reply src/routes/(console)/project-[region]-[project]/functions/function-[function]/+page.svelte]: I agree disabling is much better than hiding. Since the existing pattern hid the button, I extended the same.
I'll make the change to disable all the instances.

ItzNotABug (approved):

PR #2163, opened 2025-07-29
fix: fetch available relationship attributes

stnguyen90:
## What does this PR do?

Without this, processing/failed relationship attributes will result in 400 error.

## Test Plan

None

## Related PRs and Issues

None

Yes


ItzNotABug:
Looks like this wasn't supposed to be in `main` this early!
Should've had been reverted with `1.8.x` changes but fine I think, doesn't break!

PR #2231, opened 2025-08-19
feat: redesign custom 404 error page

HarshMN2345:
## What does this PR do?

redesigned custom 404 error page 

## Test Plan

image
image

## Related PRs and Issues

yes

### DIFF
--- a/src/routes/+error.svelte
+++ b/src/routes/+error.svelte
@@ -1,16 +1,76 @@
 <script>
     import { base } from '$app/paths';
-    import { page } from '$app/state';
-    import { Button } from '$lib/elements/forms';
     import { Unauthenticated } from '$lib/layout';
-    import { Typography } from '@appwrite.io/pink-svelte';
+    import { Button } from '$lib/elements/forms';
 </script>
 
 <Unauthenticated>
-    <Typography.Title size="xl"
-        >{'status' in page.error
-            ? page.error.status || 'Invalid Argument'
-            : 'Invalid Argument'}</Typography.Title>
-    <Typography.Title>{page.error.message}</Typography.Title>
-    <Button href={base}>Back to the console</Button>
+    <div class="content">
+        <div class="center">
+            <span>404 Page not found</span>
+        </div>
+        <h1>The page you're looking for doesn't exist.</h1>
+        <div class="center">
+            <a href={base}>
+                <Button>Back to console</Button>
+            </a>
+        </div>
+    </div>
 </Unauthenticated>
+
+<style>
+    .content {
+        text-align: center;
+    }
+
+    span {
+        padding: var(--space-1, 2px) var(--space-3, 6px);
+        border-radius: var(--border-radius-XS, 6px);
+        background: rgba(0, 0, 0, 0.08);
+        color: #56565c;
+        text-align: center;
+        font-family: var(--font-family-sansSerif, Inter), sans-serif;
+        font-size: var(--font-size-S, 14px);
+        font-style: normal;
+        font-weight: 400;
+        line-height: 140%;
+        letter-spacing: -0.063px;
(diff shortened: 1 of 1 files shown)


src/routes/+error.svelte
```diff
@@ -1,16 +1,76 @@
...
+    }
+
+    :global(.theme-dark) span {
+        background: rgba(255, 255, 255, 0.2);
+        color: #c3c3c6;
+    }
+
+    :global(.theme-dark) h1 {
+        color: #ededf0;
+    }
```
ItzNotABug: lets use classes instead of direct tags, don't want an edge case where all other tags get affected due to this global

src/routes/+error.svelte
```diff
@@ -1,16 +1,76 @@
...
 <Unauthenticated>
-    <Typography.Title size="xl"
-        >{'status' in page.error
-            ? page.error.status || 'Invalid Argument'
-            : 'Invalid Argument'}</Typography.Title>
-    <Typography.Title>{page.error.message}</Typography.Title>
-    <Button href={base}>Back to the console</Button>
+    <div class="content">
+        <div class="center">
+            <span class="error-badge">404 Page not found</span>
```
ItzNotABug: why not use `<Badge>`?

src/routes/+error.svelte
```diff
@@ -1,16 +1,76 @@
...
+        </div>
+        <h1 class="error-title">The page you're looking for doesn't exist.</h1>
```
ItzNotABug: same here for using `<Typography.*>`? is the font design different?

src/routes/+error.svelte
```diff
@@ -1,16 +1,76 @@
...
+        <div class="center">
+            <a href={base}>
+                <Button>Back to console</Button>
```
ItzNotABug: Button accepts an `href`. use a Stack for center align.

src/routes/+error.svelte
```diff
@@ -1,16 +1,76 @@
...
+    }
+
+    :global(.theme-dark) .error-badge {
+        background: rgba(255, 255, 255, 0.2);
+        color: #c3c3c6;
+    }
+
+    :global(.theme-dark) .error-title {
+        color: #ededf0;
+    }
```
ItzNotABug: we shouldn't need these if Pink2 components can be used directly.

src/lib/layout/unauthenticated.svelte
```diff
@@ -291,6 +291,13 @@
         }
     }
 
+    /* mobile only: remove top padding toeliminate extra space */
+    @media #{devices.$break1} {
+        .side-default {
+            padding-block-start: 0;
+        }
+    }
```
ItzNotABug: we don't need this. instead add the `side-default` if `!$isSmallViewport` conditionally.

src/lib/layout/unauthenticated.svelte
```diff
@@ -359,4 +366,13 @@
...
     }
+
+    .mobile-logo {
+        position: fixed;
+        bottom: 1.25rem;
+        left: 50%;
+        transform: translateX(-50%);
+        z-index: 10;
+        margin-block-start: 0 !important;
+    }
```
ItzNotABug: see - https://github.com/appwrite/console/blob/main/static/css/loading.css#L37C1-L44C2
that way, the appwrite icon doesn't shift.

src/lib/layout/unauthenticated.svelte
ItzNotABug, line 1: the default `slot` has a `u-margin-block-start-24`, lets make it conditional like -
```html
<div class:u-margin-block-start-24={!$isSmallViewport}>
    <slot />
</div>
```

PR #2423, opened 2025-09-30
Fix: loop

ItzNotABug:
## What does this PR do?

Fixes a loop inside the `$effect` block that kept firing the analytics!

## Test Plan

Manual.

Before -

Screenshot 2025-09-30 at 9 26 32 AM

After - 

Screenshot 2025-09-30 at 9 26 49 AM

## Related PRs and Issues

N/A.

Yes.

### DIFF
--- a/src/lib/components/customId.svelte
+++ b/src/lib/components/customId.svelte
@@ -27,7 +27,7 @@
             id = null;
         }
 
-        if (!id?.length) {
+        if (id !== null && !id?.length) {
             id = null;
         }
 


src/lib/components/customId.svelte
```diff
@@ -27,7 +27,7 @@
             id = null;
         }
 
-        if (!id?.length) {
+        if (id !== null && !id.length) {
```
stnguyen90, line 30: Wouldn't this effect block trigger every time `id` changes so `trackEven()` would be called every time `id` changes?

ItzNotABug: ooh yeah, missed it. fixed!

stnguyen90 (approved):

PR #2510, opened 2025-10-26
fix: correct payment method expiration notification logic

HarshMN2345:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Payment expiration warning now triggers precisely one month before a card’s expiry, avoiding premature or duplicate alerts.
  * Existing expired-card notification behavior remains unchanged and is still shown only when a card is actually expired.

### DIFF
--- a/src/lib/stores/billing.ts
+++ b/src/lib/stores/billing.ts
@@ -488,6 +488,9 @@ export async function paymentExpired(org: Organization) {
     const nots = get(notifications);
     const expiredNotification = nots.some((n) => n.message === expiredMessage);
     const expiringNotification = nots.some((n) => n.message === expiringMessage);
+    const cardExpiry = new Date(payment.expiryYear, payment.expiryMonth, 1);
+    const nextMonth = new Date(year, month + 1, 1);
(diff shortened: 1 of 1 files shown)


src/lib/stores/billing.ts
```diff
@@ -517,6 +521,26 @@ export async function paymentExpired(org: Organization) {
...
+    } else if (
+        !expiredNotification &&
+        !payment.expired &&
+        payment.expiryYear <= year &&
+        payment.expiryMonth < month
+    ) {
+        addNotification({
+            type: 'error',
+            isHtml: true,
+            timeout: 0,
```
ItzNotABug: I am not sure what the behaviour for this would be. could never hide or never show.

ItzNotABug (changes requested):

src/lib/stores/billing.ts
HarshMN2345: image
since this condition is already present so i have removed the below one which is redundant 

(The notification stays visible until the user takes action)

PR #2562, opened 2025-11-05
Fix destructive action color

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -123,27 +123,23 @@
     $: {
         if (browser) {
             const isCloudClass = isCloud ? 'is-cloud' : '';
-            const isStudioClass = PUBLIC_CONSOLE_PROFILE === 'studio' ? 'is-studio' : '';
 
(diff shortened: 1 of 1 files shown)


src/routes/+layout.svelte
```diff
@@ -314,4 +318,11 @@
             background-color: var(--color-mid-neutral-70, #56565c);
         }
     }
+
+    /* Studio profile: Override danger button text color to white */
+    .is-studio {
+        button.danger {
+            --button-color: var(--neutral-0);
```
ItzNotABug: because changing the color on theme side would mess the other text colors.

TorstenDittmann: lets solve it with just css vars - this will get out of hand here otherwise. we can make it configurable via css vars on pink

TorstenDittmann: we could also add a `profile-xxxx` class and put all possible overrides in a css file instead of here

PR #2584, opened 2025-11-11
Add Sentry env vars and refactor Sentry init

TorstenDittmann:


src/routes/+layout.ts
```diff
@@ -23,6 +24,8 @@ export const load: LayoutLoad = async ({ depends, url, route }) => {
         .then((response) => [response, null])
         .catch((error) => [null, error])) as [Account, AppwriteException];
 
+    if (account) identify(account.$id);
```
suhasdeshpande, line 28: Let's also identify it by sessionId 

```ts
const [account, error] = (await sdk.account
  .get()
  .then((response) => [response, null])
  .catch((error) => [null, error])) as [Account, AppwriteException];

if (account) {
  identify(account.$id);
  
  // Add session context to match backend
  try {
    const session = await sdk.account.getSession('current');
    identifySession(session.$id);
  } catch (err) {
    console.warn('Could not get session for Sentry', err);
  }
}
```

src/lib/sentry.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as Sentry from '@sentry/sveltekit';
+import { isCloud } from './system';
+
+export function identify(userId: string) {
+    if (isCloud)
+        Sentry.setUser({
+            id: userId
+        });
+}
+
```
suhasdeshpande, line 21:
```ts
export function identifySession(sessionId: string) {
  Sentry.setContext("Imagine Request Context", {
    session_id: sessionId
  });
}
```

PR #2601, opened 2025-11-15
fix: prevent deletion of all deployments with pending builds

woustachemax:
Fixed an issue where attempting to delete all deployments (including those still building/processing) would result in 404 errors and orphaned functions.

Changes:
- Added validation to block deletion when all deployments are selected and any are in 'waiting', 'processing', or 'building' status
- Returns proper error message to display in UI instead of allowing the operation to proceed
- Maintains ability to delete all completed deployments (leaves function in clean "no deployments" state)
- Preserves error tracking for analytics

This prevents the edge case where deleting all deployments while builds are in progress would corrupt the function state.

Screenshot from 2025-11-14 21-26-20
Screenshot from 2025-11-14 21-25-48

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented accidental deletion of all deployments when selected deployments are still processing or pending.

### DIFF
--- a/appwrite/docker-compose.yml
+++ b/appwrite/docker-compose.yml
@@ -1,0 +1,973 @@
+x-logging: &x-logging
+  logging:
+    driver: 'json-file'
+    options:
+      max-file: '5'
+      max-size: '10m'
+services:
+  traefik:
+    image: traefik:2.11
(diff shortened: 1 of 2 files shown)


eldadfux:

PR #2614, opened 2025-11-19
feat: add image transformations configuration to bucket settings

ChiragAgg5k:
## Summary
This PR adds support for configuring image transformations at the bucket level through the settings page. Users can now enable or disable image transformations for individual buckets.

## Changes
- ✨ Added transformations toggle switch in bucket settings UI
- 🔧 Updated `updateBucket` function to include `transformations` parameter  
- 📊 Added `BucketUpdateTransformations` analytics event for tracking
- 📝 Added descriptive text explaining image transformation capabilities

## Features
The new "Image transformations" section allows users to:
- Enable/disable on-the-fly image manipulation through the Appwrite API
- Control image operations like resizing, cropping, and format conversion at the bucket level
- See clear descriptions of what transformations provide

## Implementation Details
- Located between Compression and Maximum File Size sections for logical flow
- Follows the same UI pattern as other bucket settings (CardGrid with Selector.Switch)
- Respects billing read-only mode with grace period override
- Update button is disabled when no changes are made

## Test Plan
- [ ] Verify transformation toggle appears in bucket settings
- [ ] Test enabling transformations and confirm API call succeeds
- [ ] Test disabling transformations and confirm API call succeeds
- [ ] Verify analytics event fires when updating
- [ ] Confirm read-only mode prevents updates (unless grace period override)
- [ ] Check that the setting persists after page reload

## Summary by CodeRabbit

* **New Features**
  * Image transformations UI in bucket settings with toggle and standalone update action.

* **Improvements**
  * Template deployments for functions and sites now use an explicit tag + reference payload for Git-based deployments.
  * CSV import flow updated to a standardized import call.

* **Analytics**
  * Added tracking for bucket transformations updates.

* **Chores**
  * Updated package source configuration and CI workflow trigger/structure.

### DIFF
--- a/src/lib/sdk/billing.ts
+++ b/src/lib/sdk/billing.ts
@@ -490,7 +490,7 @@ export class Billing {
         name: string,
         billingPlan: string,
         paymentMethodId: string,
-        billingAddressId: string = null,
+        billingAddressId: string = undefined,
         couponId: string = null,
         invites: Array<string> = [],
         budget: number = undefined,
@@ -628,6 +628,7 @@ export class Billing {
             budget,
             taxId
         };
+
         const uri = new URL(this.client.config.endpoint + path);
         return await this.client.call(
             'patch',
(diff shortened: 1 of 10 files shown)


bun.lock
ItzNotABug, line 1: we don't use bun here, yet.

ItzNotABug (changes requested):

PR #2778, opened 2026-01-09
Fix relationship required in editor and allow point edits

kylewhirl:
## Summary
- Make relationship inputs only required when explicitly set to true to avoid forcing null relationships
- Allow spreadsheet point/linestring/polygon inline edits to parse JSON safely while typing

## Testing
- Local dev server
- Verified relationship field no longer blocks edits when null
- Verified spreadsheet inline edit for null point accepts and saves

Closes [#11111](https://github.com/appwrite/appwrite/issues/11111)

## Summary by CodeRabbit

* **Bug Fixes**
  * Relationship column validation now enforces "required" only when the column is marked required in the database, so forms reflect column metadata.
  * Spatial/string value handling is more tolerant: empty or malformed spatial inputs are handled gracefully and no longer crash or overwrite state during editing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/databases/database-[database]/table-[table]/rows/columns/types/relationship.svelte
+++ b/src/routes/(console)/project-[region]-[project]/databases/database-[database]/table-[table]/rows/columns/types/relationship.svelte
@@ -297,7 +297,7 @@
                     <Layout.Stack direction="row">
                         <InputSelect
                             {id}
-                            required={column.required === true}
+                            required={column.required}
                             {options}
                             bind:value={relatedList[0]}
                             placeholder={`Select ${column.key}`}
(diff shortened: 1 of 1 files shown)


src/routes/(console)/project-[region]-[project]/databases/database-[database]/table-[table]/rows/columns/types/relationship.svelte
```diff
@@ -297,7 +297,7 @@
                     <Layout.Stack direction="row">
                         <InputSelect
                             {id}
-                            required
+                            required={column.required === true}
```
Comment (ItzNotABug) [inline src/routes/(console)/project-[region]-[project]/databases/database-[database]/table-[table]/rows/columns/types/relationship.svelte]: just `required={column.required}` is fine.

PR #2785, opened 2026-01-14
Use `domain.verified`

hmacr:
Closes SER-1062


pnpm-lock.yaml
```diff
@@ -2597,6 +2600,9 @@ packages:
     engines: {node: '>=6'}
     hasBin: true
 
+  json-bigint@1.0.0:
```
hmacr, line 2603: `bigint` support in SDK is causing issues across Console - checking with Chirag.

PR #2805, opened 2026-01-22
Add retry button in ViewLogs modal

hmacr:
Closes SER-1069

Screenshot 2026-01-22 at 9 17 39 PM

## Summary by CodeRabbit

* **New Features**
  * Modal accepts an optional domain list and shows a Retry action for unverified domains.

* **Improvements**
  * Retry workflow to re-attempt domain verification (including optional nameserver updates on cloud), refresh status, close modal, and show success notifications.
  * Added error state, clearer retry feedback, conditional footer (only for unverified rules), and telemetry for verification attempts.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/lib/components/domains/viewLogsModal.svelte
+++ b/src/lib/components/domains/viewLogsModal.svelte
@@ -53,9 +53,7 @@
             });
             trackEvent(Submit.DomainUpdateVerification);
         } catch (e) {
-            error =
-                e.message ??
-                'Domain verification failed. Please check your domain settings or try again later';
+            error = e.message;
             trackError(e, Submit.DomainUpdateVerification);
         }
     }


src/lib/components/domains/viewLogsModal.svelte
```diff
@@ -2,17 +2,82 @@
...
+            await invalidate(Dependencies.DOMAINS);
+            show = false;
+            addNotification({
+                type: 'success',
+                message: 'Domain verified successfully'
+            });
+            trackEvent(Submit.DomainUpdateVerification);
+        } catch (e) {
+            error =
+                e.message ??
```
ItzNotABug: ideally we don't need a fallback for error messages. `e` is always `AppwriteException` when using Appwrite SDK. What was the usecase for the fallback? Was there any other `error` instance that didn't have `message` property?

hmacr: No specific use cases, just maintaining consistency with some existing patterns.
But yes, fallback isn't needed ideally - updated PR 👍

ItzNotABug: > No specific use cases, just maintaining consistency with some existing patterns.

We might need to remove those in other places too then. If the error is missing message property, something is wrong either with the SDK or the backend 👍

PR #2867, opened 2026-02-20
Fix generic invalid file error in site manual upload

ChiragAgg5k:
## Summary
- infer invalid upload reason when the dropzone emits a native picker event without detail.reason
- map inferred reason to existing extension/size notifications
- use dynamic max-size text in the size error message

## Why
Manual file picker validation can dispatch an invalid event without a reason, which currently falls through to a generic Invalid file message even for extension/size failures.

## Testing
- bun run format ✅
- bun run check ❌ (pre-existing unrelated TypeScript issues in src/lib/actions/portal.ts, src/lib/actions/tooltip.ts, vite.config.ts, warning in src/lib/layout/wizardSecondaryFooter.svelte)
- bun run lint ✅ (warnings only, no errors)
- bun run test:unit ✅
- bun run build ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file upload validation during manual site creation: now infers failure reason (size vs. format) when unspecified.
  * File size error messages now show the actual maximum limit dynamically instead of a fixed value.

* **Chores**
  * Added dependency override entries to the project configuration to lock specific package versions for builds.

### DIFF
--- a/src/lib/actions/portal.ts
+++ b/src/lib/actions/portal.ts
@@ -3,7 +3,7 @@ import type { Action } from 'svelte/action';
 
 export type PortalConfig = string | HTMLElement | undefined;
 
-export const portal: Action<HTMLElement, PortalConfig> = (el, target = 'body') => {
+export const portal: Action<HTMLElement, PortalConfig> = (el, target: PortalConfig = 'body') => {
     let targetEl;
 
     async function update(newTarget: PortalConfig) {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -88,6 +88,7 @@
         "vitest": "^3.2.4"
     },
     "overrides": {
-        "vite": "npm:rolldown-vite@latest"
+        "vite": "npm:rolldown-vite@latest",
+        "minimatch": "10.2.1"
```
HarshMN2345, line 92: @TorstenDittmann is adding the minimatch override the correct way to fix the current build failure?

PR #2886, opened 2026-02-25
fix: update site/function execution duration to not count while waiting

stnguyen90:
## What does this PR do?

We don't charge people for executions in waiting so having that timer would just make people worry.

The final duration does not include the waiting time so if it's waiting for 5s and then execution takes 20ms, the user will see the duration go from 5s to 20ms which is confusing.

## Test Plan

Manual

## Related PRs and Issues

None

Yes


eldadfux:

PR #2933, opened 2026-03-25
Enhance rule_not_found page with clearer domain error message

Srvankit:
- Updated the fallback "rule_not_found" page with a clearer and more informative message
- Clarifies that the domain is not configured in Appwrite
- Improves user experience by providing better guidance

## Summary by CodeRabbit

* **Bug Fixes**
  * Error page now shows clearer domain-configuration guidance instead of a generic 404 when applicable; otherwise it displays the underlying error message for more accurate troubleshooting.

### DIFF
--- a/src/routes/+error.svelte
+++ b/src/routes/+error.svelte
@@ -9,7 +9,7 @@
     <Layout.Stack gap="l" alignItems="center">
         <Badge variant="secondary" content="404 Page not found" />
         <Typography.Title size="l" align="center">
-            The page you're looking for doesn't exist.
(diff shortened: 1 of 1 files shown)


Srvankit:
Thanks for the feedback! That makes sense — I’ll update the implementation to ensure the message is shown only for rule_not_found cases and fallback to the default message otherwise.

Srvankit:
Thanks for the feedback! I've updated the implementation to conditionally show the domain-specific message only for rule_not_found errors, while preserving the default behavior for other error types.

eldadfux:

PR #2973, opened 2026-04-11
fix(billing): skip cloud-only organization queries on self-hosted

### LINKED ISSUE #2835: Self-hosted: 400 error on /v1/teams — unguarded billing.listOrganization() calls query non-existent 'platform' attribute
## Bug Description

On **self-hosted Appwrite 1.8.1** with **Console 7.5.7**, the organizations page (`/console/account/organizations`) returns a **400 Bad Request** error:

```
AppwriteException: Invalid `queries` param: Invalid query: Attribute not found in schema: platform
```

The console sends:
```
GET /v1/teams?queries[0]=offset(0)&queries[1]=limit(6)&queries[2]=orderDesc("")&queries[3]=equal("platform","appwrite")
→ 400 Bad Request
```

The `platform` attribute does not exist in the self-hosted teams collection schema.

## Root Cause

Two functions in `BrayF0lC.js` (billing utilities chunk) call `billing.listOrganization()` **without checking `isCloud`**:

1. **`XL()` — missingPaymentMethod check**
```js
// No isCloud guard!
async function XL() {
  const e = await sdk.billing.listOrganization([
    Query.notEqual("billingPlan", FREE),
    Query.isNull("paymentMethodId"),
    Query.isNull("backupPaymentMethodId"),
    Query.equal("platform", Platform.Appwrite)  // ← fails on self-hosted
  ]);
  // ...shows missingPaymentMethod banner
}
```

2. **`JL()` — newDevUpgradePro check**
```js
// No isCloud guard!
async function JL(e) {
  (await sdk.billing.listOrganization([
    Query.notEqual("billingPlan", FREE)  // ← also fails (billingPlan doesn't exist)
  ]))?.total
  // ...
}
```

**Most other billing functions correctly guard with `isCloud`**, for example `HL()`:
```js
async function HL(e, t) {
  if (!isCloud || !e) return;  // ✅ Correct guard
  // ...billing logic
}
```

The ternary patterns on other pages also correctly guard:
```js
isCloud 
  ? await sdk.billing.listOrganization([Query.equal("platform", Platform.Appwrite)]) 
  : await sdk.teams.list()  // ✅ Correct fallback
```

## Environment

- **Appwrite Server**: 1.8.1 (Docker: `appwrite/appwrite:1.8.1`)
- **Appwrite Console**: 7.5.7 (Docker: `appwrite/console:7.5.7`)
- **PUBLIC_CONSOLE_MODE**: `self-hosted` (correctly set in env.js)
- **Browser**: Chrome 144 (macOS)

## Expected Behavior

On self-hosted installations, `XL()` and `JL()` should either:
- Check `isCloud` before calling `billing.listOrganization()`, or
- Be wrapped in a try/catch that silently handles the failure

## Workaround

Added `'platform'` to the `ALLOWED_ATTRIBUTES` array in `Teams.php` query validator, added `platform` column + index to the MariaDB `_console_teams` table, and added the attribute to the PHP collections config.

## Suggested Fix

Add `isCloud` guard to both functions, matching the pattern already used in `HL()`:

```js
async function XL() {
  if (!isCloud) return;  // ← Add this
  // ...existing billing logic
}

async function JL(e) {
  if (!isCloud) return;  // ← Add this
  // ...existing billing logic
}
```

## Reproduction Steps

1. Install self-hosted Appwrite 1.8.1 with Console 7.5.7
2. Set `PUBLIC_CONSOLE_MODE=self-hosted`
3. Log into the console
4. Navigate to Account → Organizations
5. Open browser DevTools → Network tab
6. See 400 error on `/v1/teams` with `equal("platform","appwrite")` query

singhvishalkr:
﻿## Problem

On self-hosted Appwrite instances, the billing store calls ``organizations.list`` with cloud-only query filters (``Query.equal('platform', Platform.Appwrite)`` and ``Query.notEqual('billingPlan', ...)``). These queries return 400 errors because the self-hosted API does not support billing-related fields.

## Solution

Added early ``if (!isCloud) return`` guards at the start of ``checkForMissingPaymentMethod`` and ``checkForNewDevUpgradePro`` in ``src/lib/stores/billing.ts``. The ``isCloud`` flag was already imported from ``$lib/system``.

## How to test

1. Run Appwrite in self-hosted mode
2. Navigate to the Console
3. Verify no 400 errors from ``/v1/teams`` in the network tab

Fixes #2835

### DIFF
--- a/src/lib/stores/billing.ts
+++ b/src/lib/stores/billing.ts
@@ -586,6 +586,8 @@ export function checkForMarkedForDeletion(org: Models.Organization) {
 }
 
 export async function checkForMissingPaymentMethod() {
+    if (!isCloud) return;
+
     const starterPlan = getBasePlanFromGroup(BillingPlanGroup.Starter);
     if (!starterPlan?.$id) {
         return;
@@ -613,6 +615,8 @@ export async function checkForMissingPaymentMethod() {
 
 // Display upgrade banner for new users after 1 week for 30 days
 export async function checkForNewDevUpgradePro(org: Models.Organization) {
+    if (!isCloud) return;
+
     // browser or plan check.
     if (!browser || !org.billingPlanDetails.supportsCredits) return;
 


singhvishalkr:
Gentle nudge — also ~11 days sitting green. This one guards the organization-billing request pipeline so that on self-hosted (`isCloud === false`) we don't fire the cloud-only `/organizations/:orgId` queries that always 404 and pollute the console. Happy to tighten the check or adjust scope if the fix should live somewhere else.

fusuyfusuy:
this bug makes selfhosted deployments unusable. it tracks back to june 2025 and still no fix is merged.

eldadfux:

PR #2983, opened 2026-04-20
feat: Implement bulk delete for function deployments using Svelte 5 runes

Root-Aamir:
## What does this PR do?
This PR implements the multi-selection and bulk delete functionality for the function deployments table in the Appwrite Console. Key changes include:
* Migrated the deployments table logic to **Svelte 5 Runes** ($state, $props).
* Integrated MultiSelectionTable to enable checkboxes for selecting multiple deployments.
* Added a bulk delete action with a confirmation modal.
* Fixed TypeScript binding issues with selectedIds by using a manual selection change handler.

## Test Plan
1. Navigate to any Function > Deployments tab.
2. Select multiple deployments using the checkboxes.
3. Click the 'Delete' button in the bulk action bar.
4. Confirm deletion in the modal.
5. Verify that the selected deployments are removed and the list updates automatically.

## Related Issues
Closes #7772
232f8a5d-23be-4953-9b31-e5712176a00c
bf38fe62-20e3-43b3-bda5-064b71c6038e

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12,8 +12,8 @@ importers:
         specifier: ^1.1.24
         version: 1.1.24(svelte@5.55.0)(zod@3.25.76)
       '@appwrite.io/console':
-        specifier: https://pkg.vc/-/@appwrite/@appwrite.io/console@67539a6
-        version: https://pkg.vc/-/@appwrite/@appwrite.io/console@67539a6
+        specifier: https://pkg.vc/-/@appwrite/@appwrite.io/console@f063676
+        version: https://pkg.vc/-/@appwrite/@appwrite.io/console@f063676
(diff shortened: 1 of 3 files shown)


Root-Aamir:
"Updated! I've restored the Cancel action, added back the Redeploy guards (with tooltips), and fixed the missing column rendering for source and build sizes. I've also tightened the types for selectedIds. Ready for another look!"

Root-Aamir:
Done! I've implemented the conditional rendering for Delete and Cancel actions based on the deployment's effective status.

Root-Aamir:
"Great catch! I've moved batchDelete inside the try block. This ensures that invalidate(Dependencies.DEPLOYMENTS) in the finally block always runs to keep the UI in sync, and any errors are properly caught and tracked. Thanks for the thorough review!"

PR #2993, opened 2026-04-23
fix(vcs): restore installation ID on project Git settings

### LINKED ISSUE #1895: 🐛 Bug Report: Installation ID missing from Git configuration
### 👟 Reproduction steps

I updated to Appwrite 1.7.3 and the console no longer shows the Git installation ID, only Owner and when it was updated.

### 👍 Expected behavior

Display the installation ID so you can copy it.

### 👎 Actual Behavior

Installation ID does not show.

When navigating to the project Settings page there is an HTTP request to `v1/vcs/installations` and the response includes a `providerInstallationId: 65776313`, but using this number when [creating a function](https://appwrite.io/docs/references/1.7.x/server-nodejs/functions#create) results in an error: `AppwriteException: Installation with the requested ID could not be found. Check to see if the ID is correct, or create the installation`.

The real installation ID is found at `https://api.mysite.com/console/project-default-12345/functions/create-function`, under "Connect Git repository" in the dropdown: 

Image

Using this string when creating a function with the API works.

### 🎲 Appwrite version

### 💻 Operating system

Linux

### 🧱 Your Environment

_No response_

singhvishalkr:
I noticed the Git configuration card on the project Settings page only showed the provider owner and the last updated time. The `v1/vcs/installations` payload still includes each installation’s Appwrite id (`$id`), and that is the same value the Functions flow uses when you pick an installation in the connect-repo dropdown, so users were stuck guessing which id to pass into the Server SDK.

I added an `Installation ID` column to that table and render it with the existing `Id` tag so you can read it and copy it in one click, matching the pattern we already use on storage and other id-heavy tables.

Fixes #1895.

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/settings/updateInstallations.svelte
+++ b/src/routes/(console)/project-[region]-[project]/settings/updateInstallations.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
     import { goto } from '$app/navigation';
     import { page } from '$app/state';
-    import { Avatar, CardGrid, PaginationInline } from '$lib/components';
+    import { Avatar, CardGrid, PaginationInline, Id } from '$lib/components';
     import { Button as FormButton } from '$lib/elements/forms';
     import { getApiEndpoint } from '$lib/stores/sdk';
     import type { Models } from '@appwrite.io/console';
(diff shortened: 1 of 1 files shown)


eldadfux:

PR #2997, opened 2026-04-23
fix(sites): keep Other as the last framework filter

### LINKED ISSUE #2895: Web Template Frameworks Order Incorrect — ‘Other’ Should Be Last
### 👟 Reproduction steps

Description:
When creating a new project in Appwrite and using the Web Template creation flow, the frameworks list in the template filtering section is not displayed in the expected order. This gives the impression that the UI is not consistent.

Steps to Reproduce:

Go to Create Project → Access the project.

Click on create site

Choose clone a template

Observe that the last framework that appears is "React Native" and the second last is "Other"

Issue screenshot:

Image

### 👍 Expected behavior

The "Other" should be the last in order

### 👎 Actual Behavior

The "React Native" is the last in order and "Other" is the second last

### 🎲 Appwrite version

Version 1.8.x

### 💻 Operating system

Linux

### 🧱 Your Environment

Browser: Microsoft edge
OS: Windows 11 with WSL2 (Ubuntu 22.04)
Node.js: v22.15.1 via NVM
Package manager: pnpm v10.28.2

I already made a fix for this issue :
https://github.com/appwrite/console/pull/2896

singhvishalkr:
Fixes #2895.

When listing templates in **Create project > Create site > Clone a template**, the Framework filter currently ends with `React Native` after `Other`, even though `Other` is meant to be the catch-all bucket at the bottom of the list.

## Root cause

`src/routes/(console)/project-[region]-[project]/sites/create-site/templates/+page.ts` sorts the framework filter against a hard-coded `frameworkOrder` array that ends with `'Other'`. The comparator pushes frameworks not in that list to the end (`aIndex === -1` -> return `1`), so any newly added framework that isn't in `frameworkOrder` ends up *after* `Other`. `React Native` is one such framework -- it has live templates but is not listed in `frameworkOrder`, so it renders at the very bottom.

## Fix

- Hoist the `Other` comparison above the indexed lookup so it always sorts last, regardless of whether or not it appears in `frameworkOrder`.
- Remove `'Other'` from `frameworkOrder` since its position is no longer encoded positionally.

Net effect: any framework not explicitly enumerated (e.g. `React Native`) sits between the explicitly ordered frameworks and `Other`, which matches the reporter's screenshot expectation.

No behaviour change for frameworks already listed in `frameworkOrder`; their order is preserved.

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/sites/create-site/templates/+page.ts
+++ b/src/routes/(console)/project-[region]-[project]/sites/create-site/templates/+page.ts
@@ -36,11 +36,13 @@ export const load = async ({ url, params }) => {
         'Lynx',
         'Angular',
         'Analog',
-        'Vite',
-        'Other'
+        'Vite'
     ];
 
     const frameworks = Array.from(frameworksSet).sort((a, b) => {
+        // 'Other' always sorts last, regardless of any other framework's position.
+        if (a === 'Other') return 1;
+        if (b === 'Other') return -1;
         const aIndex = frameworkOrder.indexOf(a);
         const bIndex = frameworkOrder.indexOf(b);
         if (aIndex === -1 && bIndex === -1) return a.localeCompare(b);


eldadfux:

Issue #3076, opened 2026-06-02
Security tab not working after updating from 1.8.1 to 1.9.0


leoslamas:
### 👟 Reproduction steps

When I click the Auth/Security tab, it doesn't open. I've upgraded from 1.8.1 to 1.9.0, and I can reproduce everytime, having an 1.8.1 installation and upgrading to 1.9.0 (following the official steps)

### 👍 Expected behavior

To open all the configuration at "Security" tab page.

### 👎 Actual Behavior

The pink loading bar appears but nothing happens, and the console shows this when I click "Security":

```
CmVFoSLP.js:1 Uncaught TypeError: Cannot read properties of null (reading 'group')
    at V7AL4vJL.js:4455:2260
    at Array.filter (<anonymous>)
    at G9 (V7AL4vJL.js:4455:2248)
    at q9 (V7AL4vJL.js:4455:2399)
    at V7AL4vJL.js:291:10045
    at Q (CmVFoSLP.js:1:29591)
    at CmVFoSLP.js:1:24240
    at Zr (CmVFoSLP.js:1:27430)
    at ei (CmVFoSLP.js:1:28480)
    at B (CmVFoSLP.js:1:23332)
```

### 🎲 Appwrite version

Different version (specify in environment)

### 💻 Operating system

Linux

### 🧱 Your Environment

Hetzner CX22. Appwrite 1.9.0 (migrated from 1.8.1)

### 👀 Have you spent some time to check if this issue has been raised before?

- [x] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [x] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

WhoamiI00:
Hey, took a look at this and opened #3077 but want to be upfront — I couldn't actually pin down the exact line your stack trace points at, so it might not be the literal fix.

Here's what I found digging through the code. The Security page hits `project.listPolicies()` and immediately does `.map((policy) => [policy.$id, policy])` on the result, so a single null entry there would blow up before anything renders. The `passwordPolicies` component also reads `.total` / `.enabled` directly on policies that `+page.ts` casts without checking they exist — which would explode too if a 1.8.1→1.9.0 migration left some policies missing. Both seem like plausible upgrade-only failure modes that wouldn't reproduce on a fresh 1.9 install.

[...]

PR #3077, opened 2026-06-03
fix(auth/security): harden against missing or null policy entries (#3076)

WhoamiI00:
## What

Hardens the load function and `passwordPolicies` component for the `/auth/security` route against `null` or missing policy entries returned by `project.listPolicies()`.

## Why

After upgrading a self-hosted instance from `1.8.1` to `1.9.0`, opening **Auth → Security** crashes with:

```
Uncaught TypeError: Cannot read properties of null (reading 'group')
    at Array.filter (<anonymous>)
```

— reported in #3076.

The current code makes two unsafe assumptions:

1. `src/routes/.../auth/security/+page.ts` calls `.map((policy) => [policy.$id, policy])` on the `policies` array, so a single `null` entry throws before the page can render.
2. `passwordPolicies.svelte` accesses `historyPolicy.total`, `dictionaryPolicy.enabled`, and `personalDataPolicy.enabled` directly, but `+page.ts` already casts the lookup result as the policy type without verifying it is defined. After a migration from `1.8.1` → `1.9.0` it is possible for some policies not to be present yet, in which case those reads throw too.

## What changed

- `+page.ts` — `.filter((policy) => policy != null && policy.$id != null)` before `.map`.
- `passwordPolicies.svelte` — typed the three policy props as `… | undefined` (to match the runtime reality), used optional chaining + nullish-coalescing defaults in `onMount` and in the `hasChanges` derived.

## Verification

- `bun run check` — 0 errors, no new warnings.
- `bun run lint` — 0 errors.
- `bun run test:unit` — 235/235 pass.
- Dev server (`bun run dev`) loads with no regression.

## Caveats

The original stack trace in #3076 (`null.group` inside `Array.filter`) does not directly map to either site changed here — it points into minified JS without source maps. None of the policy models in `@appwrite.io/console` expose a `group` field, so the exact failing call site is somewhere I could not identify from the source alone (likely a code path I missed, or inside a dependency).

This PR fixes a **same-class** null-safety bug in the same load path that is reachable from the same user action. It is defensive hardening rather than a confirmed root-cause fix. If maintainers can share an unminified stack from a reproducing environment, I'm happy to follow up with a targeted patch for the precise failing site.

Related to: #3076

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/auth/security/+page.ts
+++ b/src/routes/(console)/project-[region]-[project]/auth/security/+page.ts
@@ -27,7 +27,9 @@ export const load: PageLoad = async ({ depends, params }) => {
 
     const { policies } = await sdk.forProject(params.region, params.project).project.listPolicies();
     const policiesById = Object.fromEntries(
-        (policies as ProjectPolicy[]).map((policy) => [policy.$id, policy])
+        (policies as ProjectPolicy[])
+            .filter((policy) => policy != null && policy.$id != null)
+            .map((policy) => [policy.$id, policy])
     ) as Partial<Record<ProjectPolicyId | EmailPolicyId, ProjectPolicy>>;
 
     return {
--- a/src/routes/(console)/project-[region]-[project]/auth/security/+page.ts
+++ b/src/routes/(console)/project-[region]-[project]/auth/security/+page.ts
@@ -33,27 +33,33 @@ export const load: PageLoad = async ({ depends, params }) => {
     ) as Partial<Record<ProjectPolicyId | EmailPolicyId, ProjectPolicy>>;
 
     return {
-        membershipPrivacyPolicy: policiesById[
-            ProjectPolicyId.Membershipprivacy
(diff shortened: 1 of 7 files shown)


WhoamiI00:
Good catch, you're right — the first commit only fixed two files and left the others reading policy fields in top-level `$state`/`$derived` that would still throw if any policy was missing.

Pushed c9bfd4d. The five sibling components (`updateUsersLimit`, `updateSessionLength`, `updateSessionsLimit`, `updateMembershipPrivacy`, `sessionSecurity`) now take `policy: ... | undefined` and use `?.`/`??` defaults at the read sites — `false` for the booleans, `0` for totals and duration. I also widened the eight cast-as types in `+page.ts` so the prop types actually reflect what can show up at runtime; the three email policies still go through `getDefaultEnabledPolicy` so they stay non-nullable.

Whole page should survive a partial migration now. Check/lint/tests still clean.

eldadfux:

PR #3090, opened 2026-06-22
fix: send null for unlimited user limit

TorstenDittmann:
## What
- Send null instead of 0 when updating the project user limit to unlimited.

## Why
- The API expects null to disable the user limit.

## Testing
- bun run check


HarshMN2345 (approved):

TorstenDittmann:
@greptile-apps thanks for the review. The API expects `null` on write to disable the user limit, but the policy response still returns `0` for disabled/unlimited. Keeping the read-side `policy.total === 0` checks is intentional and matches the current response contract.

Can you re-review with that context?

PR #3110, opened 2026-07-08
Resume OAuth2 authorize requests through PAR handles across login

ChiragAgg5k:
## What does this PR do?

Fixes GitHub-login breakage for OAuth2/MCP clients by resuming authorization requests through **RFC 9126 Pushed Authorization Request (PAR)** handles instead of round-tripping the full consent URL through login.

### The problem

When an MCP client (or any OAuth2 client) sends a user to `/console/oauth2/consent` and the user has no session, the consent page redirects to `/login?redirect=<full consent URL>`. That URL carries the entire authorize request — `scope` (up to 8 KB), `authorization_details`, a client `state` of up to 2 KB, and PKCE params. When the user clicks **Sign in with GitHub**, the whole thing is embedded in the OAuth `success` URL, which the backend round-trips through GitHub's OAuth `state` — and GitHub caps state size, so the login dies or comes back truncated.

### The fix

Pairs with appwrite-labs/cloud#4660, which adds `POST /v1/oauth2/{projectId}/par`:

1. **Pre-login push** — the consent page pushes the raw authorize request server-side via `oauth2.createPAR(...)` and redirects to login with only `client_id` + `request_uri`. In the verified E2E run a **2,276-char** consent URL became a **182-char** login URL — that compact URL is what travels through GitHub's `state`.
2. **Post-login resume** — a new `request_uri` branch on the consent page verifies the session first, then calls `authorize({ clientId, requestUri })` with nothing else (the server rejects any other param alongside the handle, and consumes the single-use handle even on failing calls — so it is never dereferenced while logged out).
3. **Reload safety** — once authorize returns a grant, the URL is rewritten to `?grant_id=…` with `replaceState`, so refresh/back-forward resumes via `getGrant` instead of failing on the consumed handle.
4. **Graceful degradation** — if `createPAR` fails (older or self-hosted servers without the endpoint), the page falls back to the legacy full-URL redirect, so nothing regresses.
5. **Friendly expiry** — expired/consumed handles (600 s TTL) render "This sign-in request has expired. Return to the application and try connecting again."

### Implementation notes

- **Anonymous client for PAR**: the server serves `/par` (like `/token`) with wildcard CORS and credentials disabled — per RFC 9126 any client may push without platform registration. The console SDK defaults to credentialed fetches, which the browser rejects against that posture, so `createPAR` goes through a one-off `new Client().setCredentials('omit')`. The endpoint needs no session anyway.
- **Effect keyed on the query string** (`$derived`): after email login, `goto(redirect)` is followed by `invalidate(Dependencies.ACCOUNT)`, which replaces the `page.url` *object* without changing the query. An effect tracking `page.url` directly re-runs on that identity change, cancelling the in-flight (successful!) authorize and re-dereferencing the consumed single-use handle. Deriveds only propagate on value change, which makes the flow immune to this. Caught live during E2E verification.
- `resource` params are now forwarded with `params.getAll` (RFC 8707 repeats the key; the old code silently dropped all but the first value).
- SDK pinned to the **preview build** of the cloud PR (`pkg.vc @4b3ae78`, generated via the *Cloud Console SDK Preview* workflow dispatch). ⚠️ Re-pin to the regular build after appwrite-labs/cloud#4660 merges. That spec also renamed `Models.Locale` → `Models.CloudLocale`; the seven usages are updated here.

### Flow

```
MCP client ──authorize──▶ cloud /authorize ──302──▶ /console/oauth2/consent?client_id&scope&state&…   (huge)
                                                        │  no session
                                                        ▼
                                              oauth2.createPAR(…)  ──▶  request_uri (urn:…, 600s, single-use)
                                                        │
                                                        ▼
                              /login?redirect=/console/oauth2/consent?client_id&request_uri   (~182 chars)
                                                        │  GitHub OAuth round trip fits in state
                                                        ▼
                                    authorize({clientId, requestUri}) ──▶ grant ──▶ ?grant_id=… (replaceState)
                                                        │
                                                        ▼
                                     consent card ──approve──▶ redirect_uri?code&state (state byte-identical)
```

## Screenshots

**1. Post-PAR login redirect — the resume URL is 182 chars (was 2,276) and carries only `client_id` + `request_uri`:**

login with compact redirect

**2. Consent card fully rehydrated from the pushed request after login (scopes, RAR resource selectors — URL already rewritten to `?grant_id=…`):**

consent card

**3. Consumed/expired handle — friendly retry message:**

expired handle error

## Test plan

Verified E2E with Playwright against a local cloud stack running the PR branch (public PKCE app, MCP-style request: 6 scopes + RAR `authorization_details` + 1,790-char `state`):

- ✅ Logged-out fat URL → PAR push → **182-char** login redirect containing `request_uri`
- ✅ Email login → handle dereferenced once → consent card renders the full rehydrated request
- ✅ URL rewritten to `?grant_id=…`; page reload still renders the consent card
- ✅ Approve → callback receives `code` + **byte-identical** 1,790-char `state`
- ✅ Token exchange with the PKCE verifier pushed via PAR returns access/refresh/id tokens with correct `scope` + `authorization_details`
- ✅ Re-using the consumed `request_uri` → friendly error card
- ✅ Already-approved identity: post-login `authorize({clientId, requestUri})` returns `redirectUrl` directly (skip-consent path) and lands on the callback
- ✅ `bun run format` / `check` (0 errors) / `lint` (0 errors) / `test:unit` (235 passed) / `build`

## Related

- Depends on appwrite-labs/cloud#4660 (endpoint + SDK). Draft until that merges and the SDK pin is updated.
- Cloud-side follow-up raised during review: the server's unauthenticated consent redirect still expands a `request_uri` back into full params and consumes the handle, so PAR-native clients hit the console with long URLs and a dead handle, and the console re-pushes. Passing `client_id` + `request_uri` through unconsumed would let this page skip the re-push entirely.

### DIFF
--- a/src/routes/(public)/oauth2/consent/+page.svelte
+++ b/src/routes/(public)/oauth2/consent/+page.svelte
@@ -234,6 +234,7 @@
                 // existing approved identity) via the SDK, then render consent or
                 // redirect.
                 try {
+                    const resources = params.getAll('resource');
                     const result = await sdk.forConsole.oauth2.authorize({
                         clientId,
                         redirectUri: params.get('redirect_uri') ?? '',
@@ -246,7 +247,10 @@
                         prompt: params.get('prompt') ?? undefined,
                         maxAge: parseMaxAge(params.get('max_age')),
                         authorizationDetails: params.get('authorization_details') ?? undefined,
-                        resource: params.get('resource') ?? undefined
+                        // SDK types `resource` as string; the server accepts
+                        // a URI list (RFC 8707).
+                        resource:
+                            resources.length > 0 ? (resources as unknown as string) : undefined
                     });
                     if (cancelled) return;
                     await handleAuthorizeResult(result, loggedInAccount, clientId, false);
--- a/src/routes/(public)/oauth2/consent/+page.svelte
+++ b/src/routes/(public)/oauth2/consent/+page.svelte
@@ -33,10 +33,8 @@
         phase = outcome === 'approved' ? 'approved' : 'denied';
     }
 
-    // PAR is a public client endpoint (like /token): the server serves it with
-    // wildcard CORS and credentials disabled, so the browser rejects the
-    // console SDK's default credentialed fetch. Call it through an anonymous
-    // client that omits cookies — the endpoint needs no session anyway.
+    // /par is served like /token: wildcard CORS with credentials disabled, so
+    // the browser rejects the default client's credentialed fetch.
     function anonymousOAuth2(): Oauth2 {
         const client = new Client()
             .setEndpoint(sdk.forConsole.client.config.endpoint)
@@ -45,15 +43,10 @@
         return new Oauth2(client);
     }
 
-    // The ENTIRE query string, so the flow restarts whenever any part of the
-    // authorize request changes — not just grant_id/client_id, but also
-    // redirect_uri, scope, state, nonce, PKCE fields, prompt, max_age,
-    // authorization_details and request_uri. A $derived string (rather than
-    // reading page.url in the effect) so identity-only replacements of the
-    // URL object — e.g. login calling invalidate(ACCOUNT) right after goto —
-    // do NOT re-run the flow: a rerun would cancel an in-flight authorize
-    // and, on the PAR path, re-dereference an already-consumed single-use
-    // request_uri handle.
+    // A $derived string so identity-only replacements of page.url (e.g. login
+    // calling invalidate(ACCOUNT) right after goto) don't restart the flow —
+    // a restart would cancel an in-flight authorize and re-dereference an
+    // already-consumed single-use request_uri.
     const authorizeQuery = $derived(page.url.searchParams.toString());
 
     // Re-runs when the authorize params change (this route can stay mounted as
@@ -118,8 +111,6 @@
                 return;
             }
 
-            // Shared handling for a JSON authorize response, whether the request
-            // arrived as raw params or as a pushed request_uri handle.
             async function handleAuthorizeResult(
                 result: Models.Oauth2Authorize,
                 loggedInAccount: Models.User<Models.Preferences>,
@@ -142,10 +133,8 @@
                 }
                 if (result.grantId) {
                     if (fromRequestUri) {
-                        // The pushed request handle is single-use and now consumed.
-                        // Rewrite the URL to the grant so a reload or back/forward
-                        // navigation resumes via getGrant instead of failing on a
-                        // dead request_uri. The effect re-runs and loads the grant.
+                        // The handle is now consumed — rewrite to the grant URL so
+                        // reloads resume via getGrant instead of a dead request_uri.
                         await goto(`${base}/oauth2/consent?grant_id=${result.grantId}`, {
                             replaceState: true
                         });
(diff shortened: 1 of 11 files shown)


src/routes/(public)/oauth2/consent/+page.svelte
```diff
@@ -98,19 +118,130 @@
...
+                        requestUri
+                    });
+                    if (cancelled) return;
+                    await handleAuthorizeResult(result, loggedInAccount, clientId, true);
+                } catch (e: unknown) {
+                    if (cancelled) return;
+                    const message = (e as Error)?.message ?? '';
+                    error = message.toLowerCase().includes('request_uri')
+                        ? 'This sign-in request has expired. Return to the application and try connecting again.'
+                        : message || 'Could not start authorization.';
```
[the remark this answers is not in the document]
ChiragAgg5k: Addressed in f564a72d7 — switched to matching `e instanceof AppwriteException && e.type === 'oauth2_invalid_request'`. Went with the exception type rather than the suggested 'expired'/'not found' substrings: the server reports every dead-handle case under that type, and the most common one (an already-consumed handle) returns the message `Invalid request_uri.`, which neither substring would match. Since this authorize call sends nothing but the handle, `oauth2_invalid_request` can only mean the handle itself is unusable, so the friendly retry copy is correct for all of its variants. Verified against a local cloud build of appwrite-labs/cloud#4660.

ChiragAgg5k:
Addressed the remaining Greptile finding in 32ee3fc9d — the authenticated direct-authorize path now forwards all `resource` values via `params.getAll`, matching the PAR path (RFC 8707 repeats the key; the old call dropped everything after the first value).

src/routes/(public)/oauth2/consent/+page.svelte
```diff
@@ -33,15 +33,27 @@
         phase = outcome === 'approved' ? 'approved' : 'denied';
     }
 
+    // /par is served like /token: wildcard CORS with credentials disabled, so
+    // the browser rejects the default client's credentialed fetch.
+    function anonymousOAuth2(): Oauth2 {
```
Meldiron: Hmm, why do we need this? Isnt it true that anyone can do PAR?

ChiragAgg5k: Right — PAR is deliberately public, and that's exactly why the default client can't call it. Public OAuth2 endpoints (`/par`, same as `/token`) are served with wildcard origin, and per the CORS spec a wildcard route must respond with `Access-Control-Allow-Credentials: false` (the server's `Cors.php` even throws on wildcard + credentials). Our web SDK sends every fetch with `credentials: 'include'` (session cookies), and the browser blocks a credentialed request when the response doesn't allow credentials. So this helper is not adding auth — it's the opposite: a one-off client with `setCredentials('omit')` so the browser lets the request through. Clarified the code comment in f3e76aa62.

Meldiron: PAR is moving some of /authorize logic to separate call.
How do we need it now, and didnt need it for /authroize before?

Meldiron (approved):

HarshMN2345 (approved):

PR #3119, opened 2026-07-14
fix(region): route projects in any cloud region to their subdomain (DISCORD-15265968)

claudear:
## What / Why

Reported via Discord (DISCORD-15265968): **clicking a project in the organization view gets stuck**, while the same action works from the internal impersonation tooling. Another user saw the browser throw **CORS** for the same flow.

### Root cause

`getRegionSubdomain()` in `src/lib/helpers/apiEndpoint.ts` resolved a project's regional subdomain through a hardcoded `switch` that only knew `fra` / `syd` / `nyc` / `sfo` / `sgp` / `tor`. Any project hosted in a region **outside** that list fell through to the empty default, so `buildRegionalV1Endpoint()` left the hostname unchanged and the SDK issued requests to the apex host (or whatever region is baked into `APPWRITE_ENDPOINT`) instead of the project's **real** regional host, e.g. `blr.cloud.appwrite.io`.

Those cross-origin requests failed the CORS preflight and the navigation hung. The impersonation tooling opens the project against its correct regional endpoint directly, which is why it was unaffected.

The set of Appwrite Cloud regions is dynamic and grows over time, so a hardcoded list is inherently fragile — every new region silently breaks project navigation until the switch is updated.

### Fix

- Derive the subdomain **directly from the region id** (`` `${region}.` ``) so any current or future region routes correctly. The `default` placeholder (and empty/`undefined` values) still resolve to the apex domain.
- `stripLeadingRegionSubdomain()` now also strips the region that is about to be prepended, preventing a doubled subdomain (`blr.blr.cloud…`) when the console itself is served from that region.

## Tests

Added `src/lib/helpers/apiEndpoint.test.ts` covering:
- well-known regions still map correctly,
- **regions outside the hardcoded set** (the regression) now resolve to their own subdomain,
- `default`/empty/`undefined` → apex,
- endpoint building for apex/regional hosts and the no-double-subdomain edge case.

The new tests fail against the previous logic (unlisted region → apex → CORS) and pass with this change.

### DIFF
--- a/src/lib/helpers/apiEndpoint.ts
+++ b/src/lib/helpers/apiEndpoint.ts
@@ -1,10 +1,4 @@
 import {
-    REGION_FRA,
-    REGION_NYC,
-    REGION_SFO,
-    REGION_SGP,
-    REGION_SYD,
-    REGION_TOR,
     SUBDOMAIN_FRA,
     SUBDOMAIN_NYC,
     SUBDOMAIN_SFO,
@@ -12,6 +6,12 @@ import {
     SUBDOMAIN_SYD,
     SUBDOMAIN_TOR
 } from '$lib/constants';
+
+/**
+ * The placeholder region used for region-less projects. It maps to the apex
+ * domain (no subdomain) and must never be turned into a subdomain.
+ */
+const DEFAULT_REGION = 'default';
 
 /** Ordered list of region DNS prefixes (e.g. `fra.`) for stripping from API hostnames. */
 const REGION_SUBDOMAIN_PREFIXES: readonly string[] = [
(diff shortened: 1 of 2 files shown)


claudear:
## Fix Confidence: 80/100

The mechanism is well-supported: a hardcoded region→subdomain switch means projects in newer/unlisted regions route to the apex domain, exactly matching the reported CORS-and-stuck symptom that works in impersonation (which targets the correct regional endpoint). The fix is minimal, backward-compatible, and CI's build job (which runs bun test:unit + type-check + lint + build) passed, so the new tests pass. Confidence is not higher because I could not reproduce the exact failing region in a live environment (sandbox can't install deps / no confirmed region id from the reporter), and the issue text left open that the true fault could be in cloud/appwrite backend CORS config rather than the console.

src/lib/helpers/apiEndpoint.ts
```diff
@@ -1,10 +1,4 @@
 import {
```
ArnabChatterjee20k, line 1: @claudear very bad way of fixing it here. No requests were made out of these regions. So error is somewhere else

eldadfux:

PR #3124, opened 2026-07-17
fix(databases): clarify autoscaling max is an absolute cap (DAT-1770)

abnegate:
## What

Adds helper text to the "Maximum storage (GB)" input in the dedicated database storage autoscaling settings, clarifying that the value is the absolute total size the volume may grow to — not headroom above current storage.

**Before:** label "Maximum storage (GB)" with no helper text.

**After:** same label, with helper: "The maximum total size this database's storage can grow to. Autoscaling stops at this limit."

## Why

Fixes [DAT-1770](https://linear.app/appwrite/issue/DAT-1770) — the input read as an amount of headroom to add above current storage, when it is actually the absolute cap at which autoscaling stops.

Copy-only change, no behavior.

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/databases/database-[database]/settings/updateAutoscaling.svelte
+++ b/src/routes/(console)/project-[region]-[project]/databases/database-[database]/settings/updateAutoscaling.svelte
@@ -77,6 +77,7 @@
                         min={database.storage}
                         max={10000}
                         bind:value={maxGb}
+                        helper="The maximum total size this database's storage can grow to. Autoscaling stops at this limit."
                         required />
                 {/if}
             </ul>


abnegate:
CI note: the three failing checks (build, e2e, console-stage) are the **base branch's baseline** — `feat-dedicated-db`'s own tip runs show `Tests: failure` on commits 7c838f2/2a38007 (May 8), long before this PR. This PR is a single helper-text line on an `InputNumber`; it cannot affect build/e2e. Recommend judging it on the diff.

abnegate:
Closing per maintainer decision — the base branch (feat-dedicated-db) is baseline-red and this is a 1-line copy tweak; will be folded into the feature branch's own polish instead of carried as a standalone PR.

PR #3126, opened 2026-07-20
fix: Server return html but not Rendering

### LINKED ISSUE #3125: In the "Invite member" pop-up, an `<a>` tag appears if SMTP has not been enabled on the server.
### 👟 Reproduction steps

1. Go to 'organization -> Members'
2. Click on '+ Invite'
3. Type the email and click "Send Invite"
4. (See the Image)

Image

### 👍 Expected behavior

Image

### 👎 Actual Behavior

Image

### 🎲 Appwrite version

### 💻 Operating system

Linux

### 🧱 Your Environment

appwrite/console:8.7.5

Appwrite 1.9.5

BretRen:
## What does this PR do?

fix https://github.com/appwrite/console/issues/3125

## Test Plan

Runs without bugs.

## Related PRs and Issues

https://github.com/appwrite/console/issues/3125

Yes


src/lib/components/modal.svelte
```diff
@@ -44,14 +44,14 @@
...
-                <div bind:this={alert}>
+                <div bind:this={alert} class="alert-error">
                     <Alert.Inline
                         dismissible
                         status="warning"
                         on:dismiss={() => {
                             error = null;
                         }}>
-                        {error}
+                        {@html error}
```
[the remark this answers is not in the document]
BretRen: fixed in [e2e1c2b](https://github.com/appwrite/console/pull/3126/commits/e2e1c2b4d792c2c49a1d59cbeb82bddf85169f76)

src/lib/components/modal.svelte
```diff
@@ -76,4 +80,13 @@
         opacity: 0;
         animation: none;
     }
+
+    .alert-error :global(a) {
```
HarshMN2345, line 84: No need to add a global a selector here, you can use Link.Anchor instead

src/lib/components/modal.svelte
```diff
@@ -1,8 +1,9 @@
 <script lang="ts">
-    import { Form } from '$lib/elements/forms';
-    import { disableCommands } from '$lib/commandCenter';
     import { beforeNavigate } from '$app/navigation';
+    import { disableCommands } from '$lib/commandCenter';
+    import { Form } from '$lib/elements/forms';
     import { Alert, Layout, Modal } from '@appwrite.io/pink-svelte';
+    import DOMPurify from 'dompurify';
```
HarshMN2345, line 6: do we really need this dependency?

BretRen: Personally, I don't think it's necessary since the server-side is trustworthy, but the AI ​​suggested that adding a check would be better.

src/lib/components/modal.svelte
BretRen: However, the `<a>` tag is returned by the server, so `Link.Anchor` doesn't seem to work.

src/lib/components/modal.svelte
```diff
@@ -44,14 +45,17 @@
...
+                <div bind:this={alert} class="alert-error">
                     <Alert.Inline
                         dismissible
                         status="warning"
                         on:dismiss={() => {
                             error = null;
                         }}>
-                        {error}
+                        {@html DOMPurify.sanitize(error, {
+                            ALLOWED_TAGS: ['a'],
```
HarshMN2345, line 56: If we go with this approach, can we also enforce rel="noopener noreferrer" whenever target is present? ALLOWED_ATTR only allows the attribute; it doesn’t enforce rel.

BretRen: image
image
fixed in cb6daac

src/lib/components/modal.svelte
HarshMN2345: okay, makes sense

src/lib/components/modal.svelte
HarshMN2345: was removing target intentional? That changes the behavior to always open links in the same tab. 🤔

BretRen: I'm sorry that was a silly mistake. Thanks for correcting. Fixed

HarshMN2345 (approved):

PR #3130, opened 2026-07-21
fix(databases): remove leftover per-table Usage tab

HarshMN2345:
## Summary
- Follow-up to #3128: removes the per-table Usage tab (`table-[table]/usage`), which calls `tablesDB.getTableUsage()` — an endpoint being dropped along with the backend usage-endpoint removal, and already showing no useful data on cloud.

## Test plan
- [ ] `bun run check` passes with no new errors
- [ ] Table page no longer shows a Usage tab


HarshMN2345:
Superseded by #3134 — consolidating everything into a single PR.

PR #3131, opened 2026-07-21
fix(databases): remove leftover per-database Usage tab

HarshMN2345:
## Summary
- Follow-up to #3128/#3130: removes the per-database Usage tab (`database-[database]/usage`) and its `g u` command-center shortcut. It called `tablesDB.getUsage()`, an endpoint being dropped with the backend usage-endpoint removal.

## Test plan
- [ ] `bun run check` passes with no new errors
- [ ] Database page no longer shows a Usage tab


HarshMN2345:
Superseded by #3134 — consolidating everything into a single PR.

PR #3137, opened 2026-07-22
fix: bump immutable override to clear high-severity audit

HarshMN2345:
## Summary
- `bun audit --audit-level high` was failing CI on two high-severity `immutable` advisories (pulled in transitively via `sass`): a 32-bit trie overflow DoS and a hash-collision algorithmic complexity DoS.
- Bumped the existing `immutable` override from `^5.1.5` to `^5.1.8` to reach the patched range.
- No direct dependency changes — only the transitive override.

## Test plan
- [x] `bun audit --audit-level high` — 0 vulnerabilities
- [x] `bun install --frozen-lockfile` — succeeds
- [x] `bun run check` — 0 errors

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -94,7 +94,7 @@
     "cookie": "^0.7.0",
     "devalue": "^5.8.1",
     "flatted": "^3.4.2",
-    "immutable": "^5.1.5",
+    "immutable": "^5.1.8",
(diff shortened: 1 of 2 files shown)


HarshMN2345:
Folding this into #3136 instead of a separate PR.

PR #3141, opened 2026-07-27
fix: guard against undefined table in updateKeyName

okxint:
`Array.find()` returns `undefined` when no table matches the selected `relatedTable` id (e.g. while the table list is still loading or the value was pre-populated from a saved state). Calling `.name` on `undefined` throws a runtime error.

Added an `if (table)` guard before accessing `table.name`.


okxint:
Closing.

PR #3147, opened 2026-08-04
fix(vcs): surface invalid installation tokens instead of empty results

HarshMN2345:
## Problem

When a VCS installation's OAuth token can no longer be refreshed, the API fails every repository / branch / contents call with `general_provider_failure` (`Appwrite\Vcs\InstallationTokens`). None of the Git surfaces read that error, so a dead installation was presented as a **successful empty result**:

| Surface | What the user saw |
| --- | --- |
| `repositories.svelte` | `loadRepositories()` had no `try`/`catch` at all, so the empty state claimed the org has no repositories. Used by 7 call sites |
| `branchSelector.svelte`, `productionBranchFieldset.svelte` | No error field at all: "No branches available", and the production branch fieldset stuck on its loading state forever |
| `selectRootModal.svelte` | A spinner that never resolved, or an empty tree reading as "this repository has no directories" |
| Both `updateRepository.svelte` cards | Fell back to "no repository connected" for a repository that **is** still connected, inviting the user to reconnect the repo when the installation is the problem |
| `connectRepoModal`, the create-deployment modals | Empty repo list plus a "Missing a repository? check your permissions" hint that blames scope |

## Change

`src/lib/helpers/vcsError.ts` classifies the failure into three kinds, because the remediation genuinely differs:

- **`reconnect`** - the token is dead, only re-authorizing helps
- **`locked`** (`general_resource_locked`) - a concurrent refresh holds the lock, retrying works. Deliberately **never** offers a reconnect: the installation is fine
- **`provider`** - the provider itself failed, retry first, reconnect demoted to secondary

`src/lib/components/git/installationError.svelte` renders all three with `Alert.Inline` and the reconnect link from the existing `$lib/stores/git` helpers. Reconnecting is a redirect to the **existing** authorize URL; the callback upserts the same installation row, so it is repaired in place.

Notable details:

- **The repository cards gain an explicit fourth state** rather than overloading the existing `null` sentinel, which already means "no repository connected" and is exactly the state being confused today.
- `createGitDeploymentModal.svelte` and `(modals)/createGit.svelte` are **migrated Svelte 4 -> runes**, per AGENTS.md, since they had to be touched. The `{#await load()}` in both was replaced with an explicit effect-driven load, which also avoids the runes hazard where `{#await}` re-evaluates when a signal read inside it changes.
- Modals whose primary action creates a deployment now disable that action on a non-null kind, since the call cannot succeed.

## Notes for review

- **The message substring is the only discriminator.** `general_provider_failure` covers both a dead token and a genuine provider outage; only the backend's "Please reconnect it." wording separates them. If reworded, this degrades to the `provider` branch, which still shows a real error and still offers a reconnect.
- **GitLab reconnect is not reachable from this console.** `$lib/stores/git` exports `connectGitHub` and `connectGitea` but no `connectGitLab`, so the CTA is scoped to the providers the store supports.
- **Detection is reactive only.** `listInstallations` never refreshes tokens, so a broken installation still looks healthy in the installations table until you use it.

## Unrelated backend finding, noted in passing

Not a blocker for this PR, and not introduced by it. Recording it for whoever owns VCS.

In `appwrite`, `GitHub/Callback/Get.php` initialises the token variables to `null` and only fills them inside `if (!empty($code))`, then writes them over the stored credentials unconditionally on the update path. `code` is an optional param defaulting to `''` and is never checked before that write, and the three token attributes are `required => false, default => null` in `platform.php`, so a null write succeeds silently rather than erroring. It then fails quietly afterwards, because `isExpired(null)` returns `false`.

If GitHub can return a callback with `installation_id` and no `code` (an App without "Request user authorization (OAuth) during installation", or a `setup_action=update` repository-access reconfigure), that blanks working credentials. That is a GitHub behaviour I have not verified, so treat the reachability as unconfirmed.

Worth stressing: this is reachable today through the **existing** "Configure" / "check your permissions" links, which use the same authorize and callback flow. This PR adds another entry point to it, it does not create it. GitLab and Gitea are unaffected, since their callback requires a `code`.

## Verification

- `bun run check`: **0 errors, 87 warnings** - byte-identical to the baseline on `main`
- `vitest src/lib/helpers/vcsError.test.ts`: 7/7 pass
- `eslint` and `prettier --check` clean on every changed file

### DIFF
--- a/src/lib/components/git/branchSelector.svelte
+++ b/src/lib/components/git/branchSelector.svelte
@@ -9,28 +9,79 @@
     import { Query } from '@appwrite.io/console';
     import { sdk } from '$lib/stores/sdk';
     import { page } from '$app/state';
-    import { createEventDispatcher, hasContext, tick } from 'svelte';
+    import { createEventDispatcher, hasContext, tick, untrack } from 'svelte';
+    import { getVcsInstallationErrorKind } from '$lib/helpers/vcsError';
+    import { installation, installations } from '$lib/stores/vcs';
+    import InstallationError from './installationError.svelte';
 
-    export let value = '';
-    export let installationId: string;
-    export let repositoryId: string;
-    export let label = 'Production branch';
-    export let placeholder = 'Select branch';
+    type Props = {
+        value?: string;
+        installationId: string;
+        repositoryId: string;
+        label?: string;
+        placeholder?: string;
+        /**
+         * Set to `false` when the surrounding surface already renders its own
+         * installation alert for the same installation, so the user is not shown
+         * the same "Reconnect" warning twice. The in-list error state stays
+         * either way: the branch list must never read as "no branches".
+         */
+        showInstallationError?: boolean;
+    };
 
-    const dispatch = createEventDispatcher();
+    let {
+        value = $bindable(''),
+        installationId,
+        repositoryId,
+        label = 'Production branch',
+        placeholder = 'Select branch',
+        showInstallationError = true
+    }: Props = $props();
+
+    // Deprecated in Svelte 5 but kept deliberately: every call site listens with
+    // `on:select`, and none of them are in scope here.
+    const dispatch = createEventDispatcher<{ select: string }>();
     const inDialogGroup = hasContext('dialog-group');
 
-    let open = false;
-    let searchQuery = '';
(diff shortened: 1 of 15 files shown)


eldadfux:

PR #3148, opened 2026-08-06
fix(spreadsheet): show false/0/empty string values instead of N/A badge

### LINKED ISSUE #2146: 🐛 Bug Report: NULL and FALSE values on boolean attributes show the same when selected
### 👟 Reproduction steps

Just create a collection with a boolean attribute, and then create a document and set the boolean to false

### 👍 Expected behavior

The expected thing that it should do when you click false, is that "False" should be written when selected

### 👎 Actual Behavior

When you select false, it will appear as the value has not even been set to anything (Same thing as when it's null)

Value selected to null
Image

Value selected to False
Image

### 🎲 Appwrite version

Version 1.6.x

### 💻 Operating system

Linux

### 🧱 Your Environment

I use Appwrite 1.6 on Docker in my Windows OS

I put OS to Linux because Docker uses Linux essentially 

okxint:
## What changed

In the collection spreadsheet view, the cell renderer used a bare truthiness check (`{#if value}`) to decide whether to display the value or the N/A badge. JavaScript falsy values — `false`, `0`, `""` — are all valid attribute values but were incorrectly rendered as N/A, making them visually indistinguishable from null/missing data.

## Fix

Changed the condition to `{#if value !== null && value !== undefined}` so only truly absent values get the badge.

Fixes #2146.

### DIFF
--- a/src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/spreadsheet.svelte
+++ b/src/routes/(console)/project-[region]-[project]/databases/database-[database]/collection-[collection]/spreadsheet.svelte
@@ -809,7 +809,7 @@
                                         {/if}
                                     {:else}
                                         {@const value = document[columnId]}
-                                        {#if value}
+                                        {#if value !== null && value !== undefined}
                                             <Typography.Text truncate>{value}</Typography.Text>
                                         {:else}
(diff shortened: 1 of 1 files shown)


okxint:
Closing — contributions are better targeted at the main appwrite/appwrite repo.

PR #3155, opened 2026-08-12
fix(variables): validate variable keys before writing them

levivannoort:
## What

Adds variable key validation to the console, in every flow that writes one, and fixes three flows that handled a rejected key badly.

## Why

appwrite/appwrite#13181 makes the API reject variable keys that are not valid environment variable names (`^[A-Za-z_]\w*$`, max 255). The console had **no** key validation anywhere — only a value length check — so keys with hyphens, dots, spaces or a leading digit reached the API and returned a bare server error.

Three flows turn that error into real damage:

1. **Orphaned resources.** `create-function/*` and `create-site/*` create the function or site (and its proxy rule) *before* the variables. A rejected key aborts mid-flow and leaves a resource with no deployment.
2. **Legacy keys became uneditable.** The update modal resends the *stored* key on a value-only edit, so a key stored before the rule existed started failing on every save — the user could not even correct its value.
3. **Promote-to-global could destroy data.** The conflicting branch deletes the existing global variable before creating its replacement; if the create is rejected, the original is gone.

## Changes

**New** `src/lib/helpers/variables.ts` — `isValidVariableKey`, `getVariableKeyError`, `getVariableValueError`, `validateVariables` (+ unit tests). The existing `normalizeDetectedVariables` / `mergeVariables` are unchanged.

**Validation wired into all seven entry points:** both create modals, both update modals, the `.env` import modal, and both raw editors. The existing 8192-character value check is folded into the shared helper, so each call site keeps one check instead of two.

**Behaviour fixes:**
- `updateVariablesModal` validates the key only when it changed and omits it otherwise; `handleVariableSecret` never sends it. The endpoint treats `key` as sparse (`Update.php:107`) and the SDK drops `undefined`, so a value-only edit on a pre-existing `MY-KEY` now succeeds. `sdkUpdateVariable`'s key parameter widens to `string | undefined` in three components.
- All 8 create-function/create-site flows validate before the `create()` call, so an invalid key can no longer orphan a resource.
- Promote-to-global validates before the delete.
- Keys that fail the rule are flagged with a warning icon and tooltip in both variables tables — they are ignored at build and runtime today, silently.
- The batch writes use `allSettled` and name the keys that failed, instead of surfacing one rejection for the whole batch.

## Testing

`bun run check` (0 errors, 87 warnings — unchanged baseline), `bun run lint` (0 errors), `bun run test:unit` (243 passed), `bun run build` all pass.

`src/lib/helpers/oauth2-cimd.test.ts` fails to load locally on a missing `APPWRITE_ENDPOINT`; it arrives from 9dde99783 and is untouched here.

### DIFF
--- a/src/lib/components/variables/createVariableModal.svelte
+++ b/src/lib/components/variables/createVariableModal.svelte
@@ -8,6 +8,7 @@
     import { Link } from '$lib/elements';
     import { page } from '$app/state';
     import { IconPlus, IconX } from '@appwrite.io/pink-icons-svelte';
+    import { validateVariables } from '$lib/helpers/variables';
 
     export type ProductLabel = 'site' | 'function';
 
@@ -39,13 +40,13 @@
                 newVariables = newVariables.map((variable) => ({ ...variable, secret: true }));
             }
 
-            newVariables.forEach((variable) => {
-                if (('' + variable.value).length > 8192) {
-                    throw new Error(
-                        `Variable ${variable.key} is longer than 8192 allowed characters`
-                    );
-                }
-            });
+            const validationError = validateVariables(
+                newVariables.filter((variable) => variable.key || variable.value)
+            );
+            if (validationError) {
+                throw new Error(validationError);
+            }
+
             const updatedVariables = [...variables];
             newVariables.forEach((newVar) => {
                 if (!newVar.key) {
--- a/src/lib/components/variables/importVariablesModal.svelte
+++ b/src/lib/components/variables/importVariablesModal.svelte
@@ -7,6 +7,7 @@
     import { Icon, Layout, Selector, Tooltip, Typography, Upload } from '@appwrite.io/pink-svelte';
     import { parse } from '$lib/helpers/envfile';
     import { removeFile } from '$lib/helpers/files';
+    import { validateVariables } from '$lib/helpers/variables';
 
     export let show = false;
     export let variables: Partial<Models.Variable>[];
@@ -37,10 +38,11 @@
             }
             const entries = Object.entries(uploaded);
 
-            for (const [key, value] of entries) {
-                if (value.length > 8192) {
-                    throw new Error(`Variable ${key} is longer than 8192 allowed characters`);
-                }
+            const validationError = validateVariables(
+                entries.map(([key, value]) => ({ key, value }))
+            );
+            if (validationError) {
+                throw new Error(validationError);
             }
 
             entries
--- a/src/lib/components/variables/importVariablesModal.svelte
+++ b/src/lib/components/variables/importVariablesModal.svelte
@@ -36,7 +36,9 @@
             if (!Object.keys(uploaded).length) {
                 throw new Error('No variables found');
             }
-            const entries = Object.entries(uploaded);
+            // Drop the valueless entries first. They are never written, so an
+            // invalid key on one of them must not reject the whole file.
+            const entries = Object.entries(uploaded).filter(([, value]) => !!value);
 
             const validationError = validateVariables(
                 entries.map(([key, value]) => ({ key, value }))
@@ -45,11 +47,9 @@
                 throw new Error(validationError);
             }
 
-            entries
-                .filter(([, value]) => !!value)
-                .forEach(([key, value]) => {
-                    variables.push({ key, value, secret });
-                });
+            entries.forEach(([key, value]) => {
+                variables.push({ key, value, secret });
+            });
 
             show = false;
         } catch (e) {
--- a/src/lib/components/variables/updateVariableModal.svelte
+++ b/src/lib/components/variables/updateVariableModal.svelte
@@ -7,6 +7,7 @@
     import { Layout, Selector } from '@appwrite.io/pink-svelte';
     import { Link } from '$lib/elements';
     import { page } from '$app/state';
+    import { validateVariables } from '$lib/helpers/variables';
 
     export let show = false;
     export let selectedVar: Partial<Models.Variable>;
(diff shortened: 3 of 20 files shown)


src/lib/components/variables/createVariableModal.svelte
```diff
@@ -39,13 +40,13 @@
...
             }
 
-            newVariables.forEach((variable) => {
-                if (('' + variable.value).length > 8192) {
-                    throw new Error(
-                        `Variable ${variable.key} is longer than 8192 allowed characters`
-                    );
-                }
-            });
+            const validationError = validateVariables(
```
HarshMN2345, line 43: Wouldn’t this also validate keys that are already stored? 🤔  If a project contains a legacy key (examplews MY-KEY), updating only its value would fail here before sdkUpdateVariable(..., undefined, ...) is ever reached. Should we only validate keys for newly added variables and keep validating values for existing ones? 
cc @greptile

src/routes/(console)/project-[region]-[project]/rawVariableEditor.svelte
```diff
@@ -146,22 +147,26 @@
...
 
                     if (newValue === null) {
                         await sdkDeleteVariable(variable.$id);
                     } else if (newValue !== variable.value) {
-                        await sdkUpdateVariable(variable.$id, variable.key, newValue, false);
+                        // The key is unchanged here, so leave it out and let
+                        // the API keep the stored one.
+                        await sdkUpdateVariable(variable.$id, undefined, newValue, false);
                     }
                     delete vars[variable.key];
```
Comment (HarshMN2345) [inline src/routes/(console)/project-[region]-[project]/rawVariableEditor.svelte]: Since this now uses Promise.allSettled, could a failed update leave the key in vars until the second pass? If so, wouldn’t that cause it to be processed again and end up calling `` sdkUpdateVariable``with the key, which is what this PR is trying to avoid? Would it make sense to move delete vars[variable.key] before the awaited update? 🤔 
cc @greptile

src/lib/components/variables/updateVariableModal.svelte
```diff
@@ -20,7 +21,15 @@
         secret: selectedVar?.secret
     };
 
+    let error = '';
+
     function handleVariable() {
+        const validationError = validateVariables([pair]);
```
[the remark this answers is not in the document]
HarshMN2345: @greptile False positive here, and worth remembering: this modal is only reachable through EnvironmentVariables, which is only used in sites/create-site/configuration.svelte and functions/create-function/repository-[repository]/configuration.svelte.

HarshMN2345:
@greptile re-review

HarshMN2345 (approved):

src/routes/(console)/project-[region]-[project]/rawVariableEditor.svelte
Comment (levivannoort) [reply src/routes/(console)/project-[region]-[project]/rawVariableEditor.svelte]: Good catch — you were right, and it was worse than just the key being resent: the failed variable got a second write attempt too. Fixed in 0feeb2a62 by claiming the key before the awaits rather than after:

```js
const newValue = vars[variable.key] ?? null;

// Claim the key up front. A rejected write must not leave it
// behind for the second pass, which would retry it as a new
// variable and send the stored key along with it.
delete vars[variable.key];
```

I went with moving the `delete` up rather than `try/finally` since `newValue` is already captured in a local and nothing reads `vars[variable.key]` afterwards, so there is no state to preserve across the await. Failures are still collected by `existingResults` and named in the error summary.

One note on scope: the delete branch was already safe, since `newValue === null` means the key is absent from `vars` and the `delete` was a no-op. It was only the update branch that could leak.

src/lib/components/variables/createVariableModal.svelte
levivannoort: Checked both create modals and both update modals — no change needed here, and the split you describe is what is already implemented.

**Create modals** (`lib/components/variables/createVariableModal.svelte`, `project-[region]-[project]/createVariableModal.svelte`): `validateVariables` only ever receives `newVariables`, the rows being typed into the form. The pre-existing `variables` prop is never passed through it, so stored keys are untouched.

**Update modal** (`project-[region]-[project]/updateVariablesModal.svelte`) — this is the one reachable from the post-create variables table, and it already guards on the key having changed:

```ts
const keyChanged = pair.key !== originalKey;
const validationError = keyChanged
    ? validateVariables([pair])
    : getVariableValueError(pair.key, pair.value);
```

so a value-only edit on a legacy `MY-KEY` validates the value only, and dispatches `key: undefined` so the API keeps the stored key.

`lib/components/variables/updateVariableModal.svelte` (the singular one, which greptile pointed at) validates unconditionally, but as you noted in the other thread it is only reachable through `EnvironmentVariables` in the two create-flow configuration components, where the key is always user-entered.

levivannoort:
@greptile re-review

src/lib/components/variables/importVariablesModal.svelte
```diff
@@ -37,10 +38,11 @@
             }
             const entries = Object.entries(uploaded);
 
-            for (const [key, value] of entries) {
-                if (value.length > 8192) {
-                    throw new Error(`Variable ${key} is longer than 8192 allowed characters`);
-                }
+            const validationError = validateVariables(
+                entries.map(([key, value]) => ({ key, value }))
+            );
```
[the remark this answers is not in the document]
levivannoort: Valid — fixed in e10ae018b. A line like `MY-KEY=` parses to an entry with an empty value, which the filter discards a statement later, so it was never going to be written; validating it first rejected the whole file and took the valid variables with it.

The filter now runs before validation:

```js
// Drop the valueless entries first. They are never written, so an
// invalid key on one of them must not reject the whole file.
const entries = Object.entries(uploaded).filter(([, value]) => !!value);
```

The `No variables found` check still runs against the raw parse, so a file with no parseable lines errors as before.

levivannoort:
@greptile re-review

PR #3185, opened 2026-08-26
feat(billing): ask for a GitHub reconnect when program membership cannot be verified

lohanidamodar:
> Console half of appwrite-labs/cloud#5459. That PR adds the backend state and the two `remarks` codes this reads.

## What does this PR do?

Appwrite Cloud gives education organizations a free plan through the GitHub Student Developer Program and re-validates that membership against GitHub. Validation uses the OAuth refresh token on the user's GitHub identity — and **GitHub refresh tokens expire after six months**, while revoking the OAuth app invalidates them immediately.

When that happens the backend cannot verify membership at all. Until now the console said nothing: the check simply failed and the organization drifted on a plan nobody could confirm. Cloud #5459 makes the backend warn the owners three times and then restrict the organization; this is the console side of that.

Re-authorizing needs **no backend change** — the OAuth callback binds to whoever is currently logged in regardless of how they signed in, and overwrites the stored tokens on the existing identity. So putting the user through the GitHub OAuth flow is the entire fix.

### Changes

**`src/lib/stores/billing.ts`** — constants for the two new `remarks` codes next to `billingLimitOutstandingInvoice`, and two branches in `checkForUsageLimit`, inserted after the outstanding-invoice branch:

- **Restricted** — `status === readonly` with either new code: header alert, `readOnly.set(true)`, `return`. Mirrors the outstanding-invoice branch.
- **Warning** — organization still `active` with `programMembershipUnverifiedSince` set: header alert only. **No `return`, and `readOnly` untouched**, so the budget/limit/notification paths below run unchanged. This is the window that matters — the org is warned while access is still intact.

Two alert ids rather than one, because `headerAlert.add` is a no-op when the id already exists; a single id would pin the importance to whichever state was seen first. Restricted uses 11 (matching `TeamReadonlyAlert`); the warning uses 9, below `budgetLimit` at 10, so a genuinely blocked org still leads with the blocking banner.

**`organization-[organization]/programMembershipAlert.svelte`** (new) — modelled on `teamReadonlyAlert.svelte`, same `HeaderAlert`/`Button` shape and `hideBillingHeaderRoutes` guard, written in runes per AGENTS.md. Three mutually exclusive states, and the copy differs meaningfully between them: `program_membership_invalid` means GitHub has confirmed the account is no longer a student, so reconnecting cannot help and the button points at plans instead.

**`src/lib/helpers/github.ts`** — `reconnectGithubIdentity()`, the `createOAuth2Session` call lifted from `(guest)/education/+page.svelte` with the same `read:user` / `user:email` scopes, returning the user to where they were. Shared rather than duplicated because the scopes are what mint the refresh token, and the two entry points must not drift.

**`account/identities.svelte`** — a Reconnect action per GitHub identity. That table previously offered only delete, so there was no deliberate way for a user to refresh a stale token. `trackEvent` fires *before* the call, since the OAuth flow redirects away and never returns.

### Copy

**Warning** (active) — `type="warning"`, "Reconnect GitHub to keep your education plan":
> We can't verify the Appwrite Education Program membership for **{org}** because its GitHub connection expired or was disconnected. Reconnect GitHub to keep this organization on its current plan — if the connection isn't restored, access to resources will be restricted.

**Restricted, unverified** — `type="error"`, "Access restricted":
> We couldn't verify the Appwrite Education Program membership for **{org}** because its GitHub connection expired or was disconnected, so access to resources has been restricted. Reconnect GitHub to restore access.

**Restricted, invalid** — `type="error"`, "Access restricted":
> GitHub is no longer confirming student status for **{org}**, so this organization is no longer eligible for the Appwrite Education Program and its access to resources has been restricted. Choose a plan to restore access.

The warning deliberately names the mechanical cause and avoids implying lost eligibility — a token dies at six months through no action by the member.

## Test Plan

```
bun run check       → COMPLETED 3770 FILES 0 ERRORS 92 WARNINGS
bun run lint        → 0 errors (exit 0)
bun run build       → ✓ built, Wrote site to "build"
bun run test:unit   → Test Files 1 failed | 14 passed, Tests 268 passed
```

None of the 92 check warnings or the lint warnings are in the changed files — both counts are the pre-existing baseline. The one failing unit suite (`oauth2-cimd.test.ts`, `TypeError: Invalid URL` from an absent `.env`) was confirmed pre-existing by stashing the change and re-running: byte-identical failure, same 268 passing.

**No tests added.** `vite.config.ts` configures a `client` vitest project for `src/**/*.svelte.{test,spec}.ts`, but no such file exists in the repo — all 15 suites are pure-logic tests on helpers and stores, and `billing.ts` has no existing test file. There is no component-test pattern here to follow.

Manual QA once cloud#5459 is deployed:
1. Education org, creator's GitHub identity intact → no banner.
2. Revoke the OAuth app, run `task-billing-validate-github-student --dryRun=false` → org stays active, **warning** banner appears, Reconnect GitHub starts the OAuth flow and returns to the same page.
3. After reconnecting, re-run the task → banner clears, org still active.
4. Force `status=readonly` + `remarks=program_membership_unverified` → **error** banner, Reconnect offered.
5. Force `remarks=program_membership_invalid` → **error** banner, no GitHub action, View plans instead.
6. Account → Identities: Reconnect on a GitHub identity starts the flow.

## Known gap

`programMembershipUnverifiedSince` is not yet in the pinned `@appwrite.io/console` SDK model, so it is read through one documented structural cast in `billing.ts`, following the idiom already used for `accessedAt` and `labels` in `auth/user-[user]/`. **When cloud#5459 merges and the SDK pin is bumped, delete the cast from that one function body** — nothing else changes.

## Out of scope, worth flagging

`$readOnly` is cause-agnostic but the copy attached to it is not. `organization-[organization]/members/+page.svelte:108-110` and `header.svelte:162-169` gate the Invite button on `$readOnly` and then render *"Upgrade to add more members"* / *"You've reached the members limit for the {plan} plan"*; `+page.svelte:63-70` gates project creation the same way. A student org restricted for `program_membership_unverified` will see that plan-limit wording, which is misleading. Fixing it properly means threading a restriction reason through those components.


eldadfux:

PR #3186, opened 2026-08-26
Decode UTF-16 .env files on variable import

levivannoort:
## What does this PR do?

`File.text()` always decodes UTF-8, but `.env` files written on Windows are often UTF-16 -- PowerShell's `>` redirect defaults to it. Decoded as UTF-8, every character gains an interleaved NUL byte, so importing such a file stored keys with a NUL after every letter -- an invalid environment variable name that invalidated the Kubernetes build job manifest and failed every subsequent deployment for that resource (this is the origin of a production incident on cloud).

This adds `readEnvFile()` to `$lib/helpers/envfile`: it detects UTF-16 by BOM (`FF FE` / `FE FF`), or by the interleaved-NUL byte pattern when the BOM is missing, and decodes with the right `TextDecoder`. Both import modals (`importVariablesModal`, `uploadVariablesModal`) now use it instead of `File.text()` -- so a PowerShell-authored `.env` simply imports correctly instead of erroring on every key.

## Test plan

`src/lib/helpers/envfile.test.ts` -- 7 cases: UTF-8 (with/without BOM), UTF-16LE/BE (with/without BOM), and UTF-8 containing a genuine stray NUL staying UTF-8. All passing; lint clean.

## Related PRs and Issues

- Server-side guard refusing invalid keys at the endpoints and before build job submission: appwrite/appwrite#13360


levivannoort:
Superseded by #3187 (same change with sanitized test fixtures).

PR #3193, opened 2026-09-02
fix(console): make the new Console banner responsive and snoozeable

HarshMN2345:
Three fixes found while testing the merged banner (#3192) on production.

## 1. Broken layout at tablet widths

The row switched to a column at `isSmallViewport` (`max-width: 767px`), but `Button`'s `.is-full-width-mobile` applies at `max-width: 768px`. At exactly 768px the layout stayed a horizontal row *and* the button took the full line, crushing the copy into a ~90px column seven lines tall.

| width | before | after |
| --- | --- | --- |
| 768px banner height | 182px | **96px** |
| 768px text column | 90px | **497px** |
| 768px wrap | 7 lines | **1 line** |

Direction now keys off `isTabletViewport`, and `fullWidthMobile` is dropped so the button sizes to its content.

`newDevUpgradePro.svelte` has the identical pairing (`isSmallViewport` + `fullWidthMobile`) and the same 768px break. Not touched here to keep this scoped, but worth a follow-up.

## 2. Dismissal was permanent

Closing the banner wrote a `localStorage` key that nothing ever cleared, so a single click on the X removed the message for the whole migration. For a rollout that runs months, one stray click shouldn't be a permanent opt-out.

Now uses the existing `hideNotification` helper with a **one week cool-off and exponential backoff**, and `shouldShowNotification` to gate registration. Someone who keeps dismissing it sees it less and less; someone who mis-clicks gets it back in a week.

(Note: `newDevUpgradePro` already *stores* a timestamp on dismiss and never reads it — the intent was there, unfinished.)

## 3. Resize regression from #3192

`AlertStack` owns the shell offset for alerts inside it, but only re-measured via `ResizeObserver` and `afterNavigate`. `GradientBanner` used to compensate with its own `svelte:window on:resize`, which #3192 correctly stopped running when a stack is in charge — removing the safety net.

Measured, resizing 1440 → 375 without reload:

| | before | after |
| --- | --- | --- |
| banner height | 119px | 119px |
| `header.top` | **55.99px** (stale) | **118.79px** |
| overlaps navbar | **yes** | no |

Added the resize listener to `AlertStack`, which is now the right owner. This also covers `HeaderAlert`, which has the same gap.

## Test plan

- [x] `bun run format` / `lint` / `check` — clean
- [x] Verified at 375, 768, 1024, 1440 — no overlap, offset tracks banner height
- [x] Verified resize in both directions without reload
- [x] Verified dismissal snoozes rather than permanently hides


ArnabChatterjee20k (approved):

HarshMN2345:
@greptile ## Fixed — the fallback now escalates on the same curve

Correct catch, and it was a real inconsistency: the comment directly above said *"a week at first, doubling each time"*, while the fallback wrote a flat week with no hide count. A run of rejected writes restarted the sequence every time instead of continuing it.

Extracted into `src/lib/helpers/localSnooze.ts`, reproducing the formula from `notifications.ts` exactly:

```
expiry = now + coolOffPeriod * 3600000 * factor ** (hideCount - 1)
```

Repeated failures now escalate **1w → 2w → 4w → 8w**, and a successful write still clears the local copy, so account prefs remain the source of truth and continue to carry the cross-device count.

Reads and writes are wrapped, so private mode or a corrupt value fails towards *showing* the banner rather than hiding it forever — the safer direction for a dismissable promo.

## Proof

`src/lib/helpers/localSnooze.test.ts`:

```
✓ starts at the base cool-off
✓ doubles on each repeat instead of restarting at the base period   → [1, 2, 4, 8] weeks
✓ matches the account-prefs formula, coolOff * factor ** (hideCount - 1)
✓ holds until the deadline, then releases
✓ never hides the notification for a missing or corrupt value
```

## Full CI chain

| step | result |
| --- | --- |
| `bun audit --audit-level high` | exit 0 |
| `bun run check` | 0 errors |
| `bun run lint` | 0 errors |
| `bun run test:unit` | **284/284** (275 baseline, +9 new) |
| `bun run build` | success |

PR #3203, opened 2026-09-21
fix(oauth2): keep the consent buttons disabled while the redirect is in flight

ChiragAgg5k:
## What

The consent card re-enabled its Authorize and Cancel buttons right after assigning `window.location.href`, because the `finally` block reset the busy flag before the browser had actually left the page. The browser keeps the consent page on screen until the redirect target returns its first byte. For a Sites preview that is the whole edge callback chain plus the runtime cold start, which can take many seconds, so the user could keep pressing Authorize.

A second press hits `POST /oauth2/console/grants/:id/approve` on a grant whose code was already issued, which Cloud rejects with grant-not-found. The user then sees an error toast for a flow that is in fact succeeding.

## Fix

- Add a `redirecting` state that is set right before the web redirect is issued and folded into `busy`, so both buttons stay disabled for the whole redirect chain and cold start.
- Show the spinner and `Redirecting…` on the Authorize button while in that state.
- Clear `redirecting` on `pageshow` when the page is restored from the bfcache after a back navigation, so the card is not left stuck disabled.
- Same treatment for the reject path.

The already-consented fast path in `consent/+page.svelte` already stays in the loading phase during a web redirect, so it is unchanged.

## Test plan

- [ ] Open a protected Sites preview whose runtime is cold, authorize, and confirm the button switches to `Redirecting…` and stays disabled until the site renders
- [ ] Press back from the site and confirm the consent card is interactive again
- [ ] Cancel path shows the same behaviour

## Verification

Reproduced with Playwright against the dev server: the console API is mocked, `approve` returns a redirect to a route that never sends its first byte (a preview cold-starting behind edge), then the Authorize button is pressed once and three more times over the next ~2 seconds.

| | Before | After |
|---|---|---|
| Button after the first press | `Authorize`, enabled | `Redirecting…`, disabled |
| Approve calls sent | 4 | 1 |

| Before | After |
|---|---|
| before | after |


ChiragAgg5k:
Closing: the consent page users actually hit is served by the new console; the fix moves there.

